PR #1, opened 2015-01-07
Initial azure commandline tool interface

schaefi:
Start with a proposal for the azure-cli project including
tests based on nose, commandline parsing based on docopt
and an initial structure for implementing commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+all:
+	python setup.py build
+
+install:
+	python setup.py install
+
+tests:
+	nosetests
+
+clean:
+	find -name *.pyc | xargs rm
+	rm -rf azure_cli.egg-info
+	rm -rf build
+	rm -rf dist
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ tests:
 	nosetests
 
 clean:
-	find -name *.pyc | xargs rm
+	find -name *.pyc | xargs rm -f
 	rm -rf azure_cli.egg-info
 	rm -rf build
 	rm -rf dist
--- a/azure_cli/account.py
+++ b/azure_cli/account.py
@@ -1,0 +1,28 @@
+import ConfigParser
+import os
+
+from exceptions import *
+
+class Account:
+    default_config  = os.environ['HOME'] + '/.azure_cli/config'
+    default_account = 'default'
+
+    def __init__(self, account=default_account, filename=default_config):
+        config = ConfigParser.ConfigParser()
+        if not os.path.isfile(filename):
+            raise AzureAccountLoadFailed('no such config file %s' % filename)
+        config.read(filename)
+        if not config.has_section(account):
+            raise AzureAccountNotFound("Account %s not found" % account)
+        self.config  = config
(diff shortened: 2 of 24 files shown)


schaefi:
I have added two basic command implementations for 'help' and 'container' to prove the design and also the ability to test. So far so good but now this needs a review first. Please have in mind this is my first python command line app :) I tried to make use of common practices as far as I learned and read about them.

schaefi:
what's still missing is
* decision on output format, so far I just print the data type
* decision on logger output, so far only info and error message types exists and I just prefix them with a simple INFO or ERROR

rjschwei:
More general comments rather than direct comments on the code. This is already a lot to review for a proposal of structure.

1.) We should use a different command. The command "azure" is already taken by the node.js tools and I do not want to create a conflict or overlap. Although possible not very useful one should be able to install the node.js tools and the new Python tools in parallel. Thus, msa, maz (Microsoft Azure synonyms) or acc (Azure Cloud Cli), or mac (Microsoft Azure Cloud, or Microsoft Azure Cli), or macc (Microsoft Azure Cloud Cli) or another name should be considered.

2.) Before going to town with code the README file should contain the ideas we want to implement in textual representation. For example I think we should follow what appears to be common practice today with a "primary command" that then contains "sub commands". This would look as follows:

macc --help

Prints out a help for all the available sub commands

macc storage --help

prints out the interface to manipulate storage objects

macc compute --help

prints out all options for manipulating compute resources, create an instance etc. I used "macc" as an example this does not indicate a preference for this name although I kind of like the irony of things using a command to manipulate Microsoft resources with a command that kind of hints at a big  Microsoft rival ;) 

From my perspective this pull request represents step 2, i.e. and initial sample implementation while we have not taken step one, agree on the direction via simple documentation.

schaefi:
Hi,

>    1.) We should use a different command. The command "azure"

The command is azure-cli not azure. There are no conflicts because
the sdk already took the azure namespace and I have tested that

>    2.) Before going to town with code the README file should contain the
>    ideas we want to implement in textual representation. For example I
>    think we should follow what appears to be common practice today with a
>    "primary command" that then contains "sub commands". This would look as
>    follows:

I agree and the initial container command I provided as discussion
starting point also shows that. There is:

   azure-cli help <for-command>
   azure-cli container list
   azure-cli container --help
   
and so on. The design allows easy contribution of commands by
several people without conflicts. You can see that by looking to
the tests. This must not be a one man project thus I spent some
time to think of how to let everybody contribute

>    From my perspective this pull request represents step 2, i.e. and
>    initial sample implementation while we have not taken step one, agree
>    on the direction via simple documentation.

I will write it down now that I have some input and open another
pull request. Thanks

schaefi:
documented proposal here: https://github.com/SUSE/azure-cli/pull/2

schaefi:
hope it's ok but I'm going to merge this now as a kickoff. We are still free to redesign any parts as we like
but I'd like to continue with the image registration from the blob storage and therefore I need an initial state and imho it doesn't make sense to let this PR grow any further

PR #6, opened 2015-03-27
Restructure command call syntax due to review

schaefi:
Implementing azurectl caller syntax

```
azurectl compute storage
azurectl compute image
```


schaefi:
* Tests are not yet updated
* showing help information is not yet updated

schaefi:
completion is still broken, will merge anyway to allow others to work with the code

the completion generator needs to be rewritten because the caller syntax is now more complex

PR #7, opened 2015-04-01
Handle error conditions for publishsettings import

schaefi:
Exceptions for invalid XML and/or missing Subscription section
are captured as well es decoding errors from pkcs12/base64
decryption


rjschwei:
We should have tests for the new error conditions.

schaefi:
tests done, should be ok to merge

PR #8, opened 2015-04-01
Added --quiet option for upload

schaefi:
In addition document the available option values


rjschwei:
I think we should not use positional arguments at all, i.e. <XZ-compressed-blob> and <name> and convert them to --source and --name instead.

Also please fix the commit message it says "--quit option" but should be "--quiet option"

schaefi:
thinking twice I think I understood what you mean, will change it

schaefi:
guess it's ok to merge, I need the functionality for a scripted upload. Let's open an issue if this needs to be opened again

PR #12, opened 2015-04-28
Added azurectl compute image create command

schaefi:
This command implements the creation of an Os image
from a VHD disk image file stored in a Microsoft Azure
storage container

### DIFF
--- a/azure_cli/image.py
+++ b/azure_cli/image.py
@@ -71,6 +71,11 @@ class Image:
             raise AzureBlobServicePropertyError(
                 '%s not found in container %s' % (blob_name, container_name)
             )
+        # NOTE: The information about the media_link into the storage bucket
+        # should be provided by the BlobService SDK. By the time of writing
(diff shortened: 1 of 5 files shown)


azure_cli/image.py
```diff
@@ -53,3 +56,37 @@ def list(self):
...
+            label = name
+        try:
+            storage = BlobService(self.account_name, self.account_key)
+            blob_properties = storage.get_blob_properties(
+                container_name, blob_name
+            )
+        except Exception as e:
+            raise AzureBlobServicePropertyError(
+                '%s not found in container %s' % (blob_name, container_name)
+            )
```
rjschwei: Please add a comment #TODO: The BlobService should provide the information for the blob location

schaefi: done

schaefi:
please ignore the travis checks for this pull request, they don't have access to the .travis.yml because it was added after this pull request

PR #13, opened 2015-05-04
Add SAS URL generation for containers

bear454:


azure_cli/compute_storage_task.py
```diff
@@ -94,12 +130,38 @@ def process(self):
...
+                self.command_args['--permissions']
+            )
         elif self.command_args['upload']:
             self.__upload()
         elif self.command_args['delete']:
             self.__delete()
         else:
-            raise AzureUnknownStorageCommand(self.command_args)
+            raise AzureUnknownCommand(self.command_args)
+
```
schaefi: The AzureUnknownStorageCommand exception class can then be removed from azurectl_exceptions.py
it's used only here

schaefi:
Could you add some words about this in the man page here:

doc/man/azurectl::compute::storage::container.md

schaefi:
Other than that looks great :)

schaefi:
one thing which came to my mind, did you check if setup.py and package/spec-template contains the correct module requirement list according to the new modules you used for sas support ?

azure_cli/compute_storage_task.py
bear454: It wasn't defined in `azurectl_exceptions.py`, or I would have left it be. When I started testing exceptions, it threw an error in `nosetests` as undefined.

schaefi: right, so it was an inconsistency most probably I left behind, this pull request will fix that too which is nice

schaefi:
looks all good now, I'll merge it

PR #15, opened 2015-05-05
Added file type identification for upload source

schaefi:
The azurectl compute storage upload command will now detect the
filetype and apply the decompression only for detected types.
If no supported decompressor type is found the file is uploaded
as it is

### DIFF
--- a/doc/man/Makefile
+++ b/doc/man/Makefile
@@ -17,5 +17,8 @@ install:
 		install -m 644 $$i ${man_prefix}/man1 ;\
 	done
 
+show:
+	zcat *.gz | nroff -man | less
+
 clean:
 	rm -f *.1.gz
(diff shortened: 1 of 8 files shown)


schaefi:
The damn magic implementation we provide does not exist on other distros, I'm going to change the code to just use file that will work. Please do not yet review this

doc/man/Makefile
```diff
@@ -17,5 +17,8 @@ install:
 		install -m 644 $$i ${man_prefix}/man1 ;\
 	done
 
+show:
```
bear454: Is this related?  :stuck_out_tongue:

bear454:
Looks good to me. :+1:

doc/man/Makefile
schaefi: not really but I struggled with the man page change to look good and instead to remember the command to render the page I thought it's ok to just add the show target ;)

I still enjoy the early phase of the project which allows us to just do things :)

schaefi:
cool thanks for the review, will merge it

PR #16, opened 2015-05-05
Use python logging class

schaefi:
The python logging class offers most of the functionality
we need to log our messages. However it does not provide
a progress indicator. Thus the logging class as it is used
here is also extended by a progress indicator

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ commands:
 from cli_task import CliTask
 from azure_account import AzureAccount
 from data_collector import DataCollector
-from logger import Logger
+from logger import log
 from azurectl_exceptions import *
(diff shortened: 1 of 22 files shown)


schaefi:
be careful, when this is merged the other two open pull reques #13 and #15 will have to adapt

bear454:
So I'll supplement for #13 after merge; and we can fight over supplementing for #15 after it's merged.

bear454:
Rebased from master. Sorry for the noise.

bear454:
Okay, now you're just messing with me. :dizzy_face:

schaefi:
ah uh sorry, now it should be all set for a review

PR #19, opened 2015-05-18
Added test code coverage check

schaefi:
Based on a reference file 

```
test/unit/coverage.reference.xml a check runs
```

which compares the coverage ratio from a current nosetests coverage information with the reference data. If the code coverage decreases this is an error and coverage-check invoked by 

```
make test
```

with fail. In order to update the reference data call

```
make coverage
```


schaefi:
This implementation has still a problem:

where to store the history and keep it up to date ?

rjschwei:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

On 05/18/2015 10:40 AM, Marcus Schäfer wrote:
> This implementation has still a problem:
> 
> where to store the history and keep it up to date ?

One option would be to have tests/coverage.txt in the source tree. But
in the end this acts more like a threshold than an actual comparison.

If I understood the original idea of "coverage shold not decrease"
correctly then one would run the coverage on the master branch and
then on the pull request. If the pull request coverage is greater or
equal to the result from master and the tests pass it is acceptable if
not it is rejected.

Using a coverage.txt file is more like a threshold that says "not
below this value". The threshold approach is easier to implement and
may just be as efficient.



- -- 
rjschwei@suse.com
IRC: robjo
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBAgAGBQJVWfymAAoJEE4FgL32d2UkwuEIAMjKhvX4CEHx0eASflNOTyBn
p5BkoRklObalO2EAgIkXPJhFZPJA1P0NS+0+4yYy/D2XlUjJeS+6j2biNSNhxAgd
Xi2DqAnAVBs4revQMlVyykOsOW8td96D/pl3uwCD38vsRmjoM41rd990+fO/ggTB
dKjCPnxQHxDG/uiEoBhklRIlRE6bajsM+NnHjmOSjv0QrgaVSwZGOwAF6Ot0dZFT
ZawJzB4ZsRFSmC8nU5EJ4N50LI06htZ49krF1l4HKr632CsadjkvMAJMSo1WRF92
QiIDbnn3HnGVIvd5jMTrb59VHjJt7osfk16f3n7KUXbsSN80PZcXU+9n15b7RuI=
=EbNy
-----END PGP SIGNATURE-----

schaefi:
ok let's go with a threshold file

schaefi:
I have implemented the check against a reference file. The reference file is created by nosetests and can be updated to the current state of the code via 'make coverage'. If we increase our test code coverage we should also update the reference file in order to make sure any kind of test code decrease results in an error

As of today the overall coverage is at 85%

This is because nosetests looks at all code even methods considered to be private. imho only puplic interfaces needs testing and private code should be written in a way that it does not influence public behaviour of the application. These basics of OO design are not taken into account when nosetests calculates the coverage which is ok for me. Just to explain why we are not at 100% and most likely never will be with that way of calculating the code coverage

In any case it serves the purpose to check if someone has written a test for the submitted code which afaik is the goal we have with this

Thoughts ?

rjschwei:
Sounds reasonable to me. I do have a bit of a different view on the private method testing but am good with the current approach.

schaefi:
> I do have a bit of a different view on the private method testing

which is ok. I'm sure there should be some individual style and I don't have any objections if people write tests for class private code, unless we define 100% code coverage in the meaning of nosetests
coverage

PR #20, opened 2015-05-19
s/azure?cli/azurectl/g

bear454:
Tests pass, make makes.

I suggest running `make clean` before pulling down this commit, though.


schaefi:
looks all good to me, I will merge this now and perform some smoke tests, also building the package and see if we overlooked something. Thanks

PR #21, opened 2015-05-19
Suggested changes to the existing manpage content

bear454:


schaefi:
yes, much better writing than I can do :) thanks

Issue #24, opened 2015-05-30
The tarball(s) need to be versioned


rjschwei:
When packaging the tool having tarballs that are not versioned creates an additional complication that is not necessary. It is much easier for packagers to have versioned tarballs as it is immediately obvious in the package build directory which version is being packaged.

schaefi:
This is fixed now. In order to prevent actively closing fixed issue please do not use the indention style with [+.-] in your commit messages, and just write Fixes #id, github seems to be unable to notice the fix otherwise. Thanks

PR #25, opened 2015-05-30
Fix issue #24

rjschwei:


schaefi:
In principal I don't have a strong opinion on this, but there is one thing which bothers me with version names in source tarball. Any time you submit a new package you have to osc delete the old source and osc add the new one. I find it pretty nice if I don't have to care for these kind of name change at all and just copy the new sources over the old ones and submit the thing

schaefi:
a short look into the spec file outlines the versions too and I'm mostly interested in the version when I need to debug a customer problem which requires me to ask for rpm -q or azurectl --version anyways

rjschwei:
If the tarball i snot versioned there is nothing to, at a quick glance, confirm consistency. I can set the sec file version to 1.0 and still have the tarball version from 0.9 simply because I forgot to copy the newly downloaded 1.0 tarball into the package. Versioning the tarball provides a little extra "protection" from silly mistakes.

schaefi:
ok

PR #27, opened 2015-06-02
[Review] Request from 'schaefi' @ 'SUSE/azurectl/review_150601_azurectl_config'

schaefi:
Added setup service and account command e.g

```
azurectl setup account list
```


azurectl/account_setup.py
```diff
@@ -0,0 +1,81 @@
...
+        self.config = ConfigParser.ConfigParser()
+        self.filename = filename
+        try:
+            if os.path.isfile(filename):
+                log.info('Parsing config file: %s' % filename)
+                self.config.read(filename)
+        except Exception as e:
+            raise AzureConfigParseError(
+                'Could not parse config file: %s: %s' %
+                (type(e).__name__, format(e))
```
rjschwei: using "filename" as the first argument in the tuple makes this more obvious.

azurectl/account_setup.py
```diff
@@ -0,0 +1,81 @@
...
+        return True
+
+    def add(
+        self, name,
+        publish_settings_file, storage_account_name, container_name
+    ):
+        self.config.add_section(name)
+        self.config.set(
+            name, 'publishsettings', publish_settings_file
+        )
```
rjschwei: Should validate that the file actually exists. Whenever possible we want to avoid writing a bad config file.

rjschwei:
All methods should have a doc-string even if the method name makes this very obvious. The doc-string will help with elineation. For example the add method looks odd at first glance with a ): on a line by itself. A doc-string will help facilitate visually that the ): is part of the method declaration.

schaefi:
This reminds me to the practical OO design book from Sandy Metz, I really liked that book and applied many of the suggested design patterns to my code these days. One of it is to avoid long argument lists for methods and use a hash which makes this more flexible and readable. I don't think using doc strings to make the method definition more readable is the right way to go.

How about it now ?

Thanks

rjschwei:
This does not work for me at all, especially in Python. The "hash" as argument works equally well in Perl because Perl globs all arguments together anyway. However, the Python argument list is already a dictionary as Python supports keyword arguments.

The key problem for me with the "hash as argument" beside the idiomatic issue for Python, is also that one cannot look at the definition to understand the API. In order to know the arguments one has to always look at the whole function or the arguments have to be listed in the doc string. These are additional complications we should avoid.

There is one misunderstanding from my previous comment that I would also like to address. All functions should have a doc string. The delineation to the code from the interface definition is a nice side effect of the doc string but certainly not it's primary reason.

Long argument lists can be formatted to be reasonably visually appealing.

schaefi:
I have changed back to argument lists and added doc strings because I don't have a so strong feeling about it and can deal with both

However:

* I don't see much value in doc strings in this case. The class name and also the method names in my opinion makes it clear what the method does. Adding doc strings here just increases the problem that at some point the method header doc is not in line with what the function is good for. Therefore people tend to keep this information as generic as possible and also as short as possible so that they don't have to touch or remember it, which in my opinion makes it a pointless exercise. I like doc strings in an application to explain not obvious code to me, in situation were this doc information is important and also maintained by the people when they change that code. Because I don't want to stand in anybody's way I added the doc strings in this case but if you feel we need this for the entire application please go ahead and add them to the other parts of the code ;)

* With regards to argument lists of methods,  I think handling a hash here would have had the advantage that the caller semantic does not change if new items will be added in the future to setup an account. Especially default values could have been added without taking care in which order the information is provided. If a method is defined as:

```
def methods(self, args):
    arg_a = args[a]
    arg_b = args[b]
    arg_c = args[c]

    ...implementation
```

it still is clear to me how the interface looks like and it's more flexible for change and testing
Even a doc string imho fits better to this style

Just my 2cents

azurectl/account_setup.py
```diff
@@ -0,0 +1,105 @@
...
+    def add(
+        self, section_name, publish_settings, storage_account, storage_container
+    ):
+        """
+            add new account section
+        """
+        self.__validate_publish_settings_file(
+            publish_settings
+        )
+        self.config.add_section(section_name)
```
rjschwei: Sorry, yet another thing....

Somewhere we need to check whether the account already exists. If it does we need to complain. I am not certain if add_section on the config object throws an exception if the section name already exists or not. If it does we can use that mechanism, it it doesn't we have to check first before adding.

azurectl/account_setup.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        self.__write()
+        return True
+
+    def __validate_publish_settings_file(self, filename):
+        """
+            validate given publish settings file for
+            + existence
+        """
+        if not os.path.isfile(filename):
```
rjschwei: I think it is fair to create the file if it does not exist. Then as a user I can just create the file through te command line without having to really know where it lives etc.

azurectl/account_setup.py
schaefi: no problem that's what review is good for

You will get a ConfigParser.DuplicateSectionError but as an unexpected error which it should not be. I'll change that

azurectl/account_setup.py
schaefi: Don't understand that. You want the tool to touch an empty publish settings file if it does not exist ?

rjschwei: Sorry, my bad. Yes, we need to complain when there is no publish settings file. This is correct.

PR #28, opened 2015-06-03
Add information about signing of code commits

rjschwei:


schaefi:
looks great and also fixes style issues, will merge it now

PR #29, opened 2015-06-08
[Review] Request from 'schaefi' @ 'SUSE/azurectl/review_150608_doc_update_azurectl_config'

schaefi:
Please review the following changes:
  * b17be75 Update README
  * bdda36a Added information about azurectl setup account use

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Command Line Interface to manage
     - [Code Structure] (#code-structure)
   * [Compatibility] (#compatibility)
   * [Issues] (#issues)
-  * [Signing GIT patches] (#signing)
+  * [Signing GIT patches] (#signing-git-patches)
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,9 +93,21 @@ for a default account:
...
+
+In order to create or manage the account sections in the config file,
+azurectl provides the __setup__ command. The default account section
+as shown above could be created using the following command
+
+```
+$ azurectl setup account add \
+  --name default \
+  --publish-settings-file /path/to/publish/settings/file \
+  --storage-acainer-name storage-account-name \
```
rjschwei: I take it this should be "--storage-container-name". However I would claim this to be confusing as the next option is "--container-name". Thus I'd say "--storage-account-name" is probably the better choice.

README.md
```diff
@@ -425,51 +438,59 @@ compatibility layer.
...
-[user]
+   ```
+   [user]
        name = Joe Developer
        email = developer@foo.bar
        signingkey = 11223344
+   ```
 
-and thats basically it.
+   Thats basically it.
```
rjschwei: Should be "That's" as in "That is"

My mistake I I wrote the original ;)

schaefi:
Thanks for the review, I updated it

PR #30, opened 2015-06-24
Added support for the Azure File Share service

schaefi:
azurectl compute storage share list, create and delete will be
added with this commit. Once created a file share can be mounted
using a CIFS mount

### DIFF
--- a/.travis.script
+++ b/.travis.script
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Clone Azure SDK development fork on github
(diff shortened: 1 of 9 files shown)


schaefi:
This is based on the Azure SDK pull request I made to support File Share

schaefi:
This is based on the pending SDK pull request I made for Azure. Therefore the tests are failing because the code is not yet upstream. Please do not merge until then

PR #32, opened 2015-07-05
Added RequestResult class

schaefi:
Encapsulate operations based on a request ID returned from the
Azure API. This makes the code more explicit as well as allows
to think of commands like

    azurectl compute request status --id <id>
    azurectl compute request wait --id <id>

in the future


rjschwei:
LGTM needs to be rebased I guess as I merged #33

PR #34, opened 2015-07-15
Minor version bump.

bear454:
In accordance with [semantic versioning](http://semver.org/), both the `azure files` functions
and the `compute vm` functions should have been accompanied by a version
bump.

Time to catch up.


schaefi:
right I completely forgot about that, thanks

PR #35, opened 2015-07-16
Remove logging noise during uploads.

bear454:
The scheduler was still triggering events during upload, causing lots of
log noise around the progress bar.  Proposed solution is to add another
class of event for logging to ignore.

TODO: subversion bump on merge.


schaefi:
good catch, thanks

PR #38, opened 2015-07-16
Short switch `-v` for version.

bear454:


schaefi:
Thanks, I have tested the completion generator with this change and unfortunately it broke. I'm going to fix the completion generator first and will then merge this commit

schaefi:
done now, merging

bear454:
Sorry about the completion generation - maybe we can automate testing that? ;-)

schaefi:
>    Sorry about the completion generation - maybe we can automate testing
>    that? ;-)

also thought about it but it's pretty hard to test bash code and I'm not aware of any completion automation which provides me a good return result for a test. I think it would be possible if we rewrite the completion python code which creates the bash sequences to be testable. But that would be a not so small side project which I would consider not a prio at the moment

PR #39, opened 2015-07-21
Update code base to use SDK's dev branch

schaefi:
Microsoft will release the dev branch to be the new SDK in August. The version will be at 0.20.x and is incompatible to 0.11.x I have ported the files api support to the dev branch and also updated azurectl to work with the latest dev branch

### DIFF
--- a/.travis.script
+++ b/.travis.script
@@ -4,7 +4,7 @@
 git clone https://github.com/schaefi/azure-sdk-for-python.git
 
 # Install specified development branch
-branch_name=azure-files-api
+branch_name=dev
 (
     cd azure-sdk-for-python
     git checkout $branch_name
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ python:
   - "2.7"
 
 install:
-# The azure SDK is currently taken from a development branch
-# Once our PR (https://github.com/Azure/azure-sdk-for-python/pull/394)
-# is merged in master we can use azure from pip again
+# The azure SDK is currently taken from the development branch (dev)
+# We can switch using azure from pip once Microsoft has released
+# the new version which is planned to happen in August 2015
 #
 #  - "pip install azure"
 #
@@ -20,5 +20,6 @@ install:
   - "pip install pyliblzma"
   - "pip install coverage"
   - "pip install pyOpenSSL"
+  - "pip install requests"
 
 script: ./.travis.script
--- a/azurectl/cloud_service.py
+++ b/azurectl/cloud_service.py
@@ -15,7 +15,7 @@ import base64
 import subprocess
 from tempfile import NamedTemporaryFile
 from azure.servicemanagement import ServiceManagementService
-from azure import WindowsAzureConflictError
+from azure.common import WindowsAzureConflictError
 
 # project
 from azurectl_exceptions import *
--- a/azurectl/container.py
+++ b/azurectl/container.py
@@ -13,9 +13,9 @@
 #
 import datetime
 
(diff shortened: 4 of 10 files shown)


schaefi:
I had a good discussion with $MS on the development branch side of things. As you can see with this pull request I'm up and running with the dev branch they will release in August.

Along with this change I suggest we also rebuild our python-azure-sdk rpm package to use the latest dev branch.

I'm confident that my new files pull request will make it into the dev branch

schaefi:
According to $MS plans we should wait with our azurectl submission until August when they release 0.20.x of the SDK. imho this gives us also some time to add support for the publish and replication api

schaefi:
For testing I have build python-azure-sdk 0.20.x in Cloud:Tools

rjschwei:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

On 07/23/2015 10:32 AM, Marcus Schäfer wrote:
> Merged #39.

This was pre-mature IMHO, now we end up using an SDK that is according
to the upstream project unstable with additional API changes to come.
I would have much rather waited until 0.20.x was actually released. It
is getting harder to contribute to azurectl and the SDK by the minute :(




- -- 
rjschwei@suse.com
IRC: robjo
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

iQEcBAEBAgAGBQJVtnEHAAoJEE4FgL32d2UkBqYIAIl8PMdhvRHuU8cYLHDUTh3W
rtjUrvaxhhkRNykQMNGzai/Qj0TEQ7jeFmC70W+QTuTRy7D0n23/ppmKkdhXaD1j
6MQnodNqXf5/POGS0VnU2jwK0NKDe4x3e0AdSQ4+DEPysNwX/+8Jgvz7Lrg2j5OW
51Do9VgvfRDkULxJoh5PtdG0hFWTYrapAA7jvanUZauG6B6aZuuu1xDse2CRzZBN
PUlA6kcbxp3cidm14LoveTepHzIR2m5UqlmQs4cCS3L4OLbtv5ySo2Mt8afy/y32
uksDuMgd7CnnAWp5zvGM1P3wow4cdP82MuTRicFWmEg0iT0tyPHJVmiKxEqvIEw=
=kH/i
-----END PGP SIGNATURE-----

schaefi:
>    This was pre-mature IMHO, now we end up using an SDK that is according
>    to the upstream project unstable with additional API changes to come.

There is no other way if we want to continue developing azurectl with
support for file shares, publishing and replication. They will
not accept any patches to the stable branch

>    I would have much rather waited until 0.20.x was actually released. It
>    is getting harder to contribute to azurectl and the SDK by the minute
>    :(

There is no problem, our code is covered 100% by tests, any incompatible
change in the api is immediately recognized and we can follow their
development which we have to until they release.

The release will happen in August, this is not so far away and I want
to have an azurectl version working from day one once they will release

I agree we will not promote azurectl or put it into sles until they
have released the dev branch to be the new sdk

Support for file shares is in now in, thus I already could get rid of one
patch we carry around. For me personally things became easier as before
because I can follow their work

Bunch of the SDK code will also be autogenerated in the future.
I also discussed with them that the new sdk is incompatible to 0.11
in namespace and other parts. They know that this is a break but
due to their internal restructuring process (storage handled by another team)
they don't see another way

For me there is no other way as following their style and sticking to
the current master branch is a waste of time, nothing will work once
they do the switch.

Instead of dealing with tons of problems I wanted to be there in time
Actually I'm quite happy that I got it all running including file share
support in the sdk and not as a suse patch

It's a pity that you think differently

PR #40, opened 2015-07-23
Inserted requirements.txt for pip

tomschr:
Useful as you don't need to change `.travis.yml` and only the file `requirements.txt`.

For more information, see https://pip.readthedocs.org/en/1.1/requirements.html


schaefi:
maybe nit picking but could you rename requirements.txt to a hidden file .requirements.txt ?

It's only needed in the scope of travis and that config lives also in a hidden file

In the past I thought of using the install_requires from setup.py but it has turned out the testing env requirements needs more stuff on top of the runtime requirements. Thus we still need two places to define this information

tomschr:
> maybe nit picking but could you rename requirements.txt to a hidden file .requirements.txt ?

@schaefi Sure, no problem. See my last commit. :-) I was just followed the usual convention and thought it could be fine too.

tomschr:
> In the past I thought of using the install_requires from setup.py but it has turned out the testing env requirements needs more stuff on top of the runtime requirements. Thus we still need two places to define this information

Hmn, not sure if I understood you correctly, but you could nest requirement files. For example, in my [openSUSE/docmanager](https://github.com/openSUSE/docmanager) repository we have a `test_requirements.txt` and a `requirements.txt` file.

The `requirements.txt` contains everything we need to run docmanager whereas the `test_requirements.txt` contains just these lines to enable our testing framework:

```
-r requirements.txt
pytest
```

In that case, I don't have to repeat the basic requirements again; I just add a pointer. :grinning: 

Regarding the `install_requires` vs. pip requirements, maybe you could just read in the requirement file, create a list and pass it to the `install_requires` keyword.

You may find the following link maybe helpful; it contains an analysis of "install_requires" vs pip's
requirements files: https://packaging.python.org/en/latest/requirements.html#requirements-files

schaefi:
I was more thinking of having something like

```
config = {
   ...
   'travis_requires': [ .... ]
}
```

in setup.py and avoid having an extra requirements.txt file. But I didn't look closer if that wouldn't confuse python setuptools, and I also did not check how to move the information to be consumable in the .travis.yml. I just thought it would be nice to have one place and not several files. But all this is
just a cosmetic thing

The change you did is nice and makes the travis setup small. I'm going to merge this now

PR #43, opened 2015-08-24
Update travis tests due to Azure SDK restructuring

schaefi:
The Azure Storage part of the SDK has been moved into its own
git repository. This testing and also packaging of the SDK
has changed


schaefi:
Standard testing procedure using azure from pip also works again now since $MS has merged everything to master. I'll keep the way of using the code from github because we plan to add publishing and replication support to the SDK soon

PR #45, opened 2015-08-26
Support for VM publishing

schaefi:
Added support for publishing previously replicated images


schaefi:
I have updated the command. By default publish does what the name says it publishes but with the option --private or --msdn you can limit the scope

PR #46, opened 2015-09-09
Handle accounts with more than one subscription

bear454:
In the event an Azure account holder has more than one subscription, the publishsettings file will have many `<subscription>` tags. 

Its perfectly acceptable (and quite convenient) to leave the subscription id out if only one subscription is in use, but in the case of multiple subscriptions, there needs to be some mechanism for selecting the appropriate subscription.


schaefi:
great catch, great patch :)

Thanks

Issue #41, opened 2015-07-24
Look for an Additional Configuration File under ~/.config/azurectl/ (XDG)?


tomschr:
In the "old" days, configuration files are stored directly under `$HOME` like `~/.bashrc`. There is nothing wrong with that approach.

However, it seems thesedays more and more applications store there configuration files under `~/.config`. This directory comes from the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

Maybe `azurectl` could look to find `~/.config/azurectl/config`first . If it fails, it could try it with `~/.azurectl/config`.

Just an idea. :smile:

schaefi:
This could be certainly added. We are currently on hold with a release of azurectl to sles12 until Microsoft will release the new version of the SDK. This will happen in August. Once azurectl is publicly available we will see if there is demand for an alternative config file location

Does that sound fair to you ?

tomschr:
Sure, absolutely! :+1:  Just an idea, not all my ideas are good. :wink:

bear454:
I agree; we should change `azurectl` to default to `~/.config/azurectl/`, and since we have it, fall back to `~/.azurectl/`.  Thanks @tomschr for finding the spec for this.

rjschwei:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

On 07/27/2015 01:47 PM, James Mason wrote:
> I agree; we should change `azurectl` to default to
> `~/.config/azurectl/`, and since we have it, fall back to
> `~/.azurectl/`.  Thanks @tomschr for finding the spec for this.
> 

If we go this direction, which is fine with me I would propose that we
move existing configuration files to the new location.

Thus we would:
- - look in ~, if we find the config file
  + move it to ~/.config/azurectl/ if there is no config file
  + merge with ~/.config/azurectl/ if a config file exists
    ~ do nothing if a merge is not possible, i.e. same accounts exist
      with different credentials
  + remove the config file from ~ after successful merge
- - create new files in ~/.config/azurectl/
- - deprecate the ~ location of the config and print a message
  + remove the merge code at a given date
  + remove the code that looks at ~ at the same date as the merge code



- -- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo
-----BEGIN PGP SIGNATURE-----
Version: GnuPG v2

[...]

PR #47, opened 2015-09-17
Moved primary default config file location

schaefi:
With respect to the XDG Base Directory Specification the default
config file location for azurectl has changed. The default config
file is now searched in the following order and paths

1. ~/.config/azurectl/config
2. ~/.azurectl/config

This fixes issue #41

### DIFF
--- a/azurectl/account_setup.py
+++ b/azurectl/account_setup.py
@@ -33,7 +33,6 @@ class AccountSetup(object):
         self.filename = filename
         try:
             if os.path.isfile(filename):
-                log.info('Parsing config file: %s', filename)
                 self.config.read(filename)
         except Exception as e:
             raise AzureConfigParseError(
--- a/azurectl/config.py
+++ b/azurectl/config.py
@@ -58,7 +58,7 @@ class Config(object):
                     )
                 )
         try:
-            log.info('Using config file %s', self.config_file)
+            log.debug('Using configuration from %s', self.config_file)
             usr_config.read(self.config_file)
         except Exception as e:
             raise AzureConfigParseError(
--- a/azurectl/config.py
+++ b/azurectl/config.py
@@ -42,7 +42,7 @@ class Config(object):
             account_name = 'default'
         if filename and not os.path.isfile(filename):
             raise AzureAccountLoadFailed(
-                'no such config file %s' % filename
+                'Could not find config file: %s' % filename
             )
         elif filename:
             self.config_file = filename
--- a/azurectl/config.py
+++ b/azurectl/config.py
@@ -32,6 +32,8 @@ class Config(object):
     PLATFORM = sys.platform[:3]
 
     def __init__(self, account_name=None, filename=None, platform=PLATFORM):
+        from logger import log
+
         usr_config = ConfigParser()
         self.config_files = [
             '.config/azurectl/config', '.azurectl/config'
@@ -56,6 +58,7 @@ class Config(object):
                     )
                 )
         try:
+            log.info('Using config file %s', self.config_file)
             usr_config.read(self.config_file)
         except Exception as e:
             raise AzureConfigParseError(
--- a/azurectl/config.py
+++ b/azurectl/config.py
@@ -48,8 +48,12 @@ class Config(object):
             self.config_file = self.__default_config(platform)
             if not self.config_file:
                 raise AzureAccountLoadFailed(
-                    'no default config found, searched for %s in home dir: %s' %
-                    (':'.join(self.config_files), self.__home_path(platform))
+                    'could not find default configuration file %s %s: %s' %
+                    (
+                        ' or '.join(self.config_files),
+                        'in home directory',
+                        self.__home_path(platform)
+                    )
                 )
         try:
             usr_config.read(self.config_file)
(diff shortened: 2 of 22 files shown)


azurectl/config.py
```diff
@@ -32,24 +32,38 @@ class Config(object):
...
+            raise AzureAccountLoadFailed(
+                'no such config file %s' % filename
+            )
+        elif filename:
+            self.config_file = filename
+        else:
+            self.config_file = self.__default_config(platform)
+            if not self.config_file:
+                raise AzureAccountLoadFailed(
+                    'no default config found, searched for %s in home dir: %s' %
```
rjschwei: 'Could not find default configuration file %s in home directory: %s' % ('or'.join(self.config_files), self.__home_path(platform))

azurectl/config.py
```diff
@@ -61,12 +75,18 @@ def get_option(self, option):
...
             else:
                 homeEnvVar = 'UserProfile'
-        config_file_path = os.environ[homeEnvVar] + '/.azurectl/config'
-        return config_file_path
+        return os.environ[homeEnvVar]
+
+    def __default_config(self, platform):
+        for filename in self.config_files:
+            full_qualified_config = self.__home_path(platform) + '/' + filename
+            if os.path.isfile(full_qualified_config):
```
rjschwei: if not '.config' in full_qualified_config:
  print 'WARNING config file location %s has been deprecated move the config file to ~/.config/azurectl/'

schaefi: Hmm, this one I don't like. Either we support two locations for the config file and both are ok to use or we only support one location. I'm not in favour to carry a warning message which imho stands in my way when using the tool.

actually I like to support both locations, there are old school linux users who like ~/.azurectl and there are the modern linux users who like to have all under .config. The current implementation makes both happy. Adding a warning message makes one group feel bad

For me there is no disadvantage to support two locations for the same file and give .config the higher priority.

I would say it's more important to add a message which config file will be actually used. I'm going to add that

Thoughts ?

schaefi:
Sorry the patch became a little longer because I realized we read in the config file multiple times.
The patch for this is in a single commit if you prefer to review on a commit base instead of the full change

azurectl/config.py
rjschwei: On 09/18/2015 03:05 AM, Marcus Schäfer wrote:
>>          homeEnvVar = 'HOME'
>>          if platform == 'win':
>>              if 'HOMEPATH' in os.environ:
>>                  homeEnvVar = 'HOMEPATH'
>>              else:
>>                  homeEnvVar = 'UserProfile'
>> -        config_file_path = os.environ[homeEnvVar] + '/.azurectl/config'
>> -        return config_file_path
>> +        return os.environ[homeEnvVar]
>> +
>> +    def __default_config(self, platform):
>> +        for filename in self.config_files:
>> +            full_qualified_config = self.__home_path(platform) + '/' + filename
>> +            if os.path.isfile(full_qualified_config):
> 
> Hmm, this one I don't like. Either we support two locations for the config file and both are ok to use or we only support one location. I'm not in favour to carry a warning message which imho stands in my way when using the tool.
> 
> actually I like to support both locations, there are old school linux users who like ~/.azurectl and there are the modern linux users who like to have all under .config. The current implementation makes both happy. Adding a warning message makes one group feel bad
> 
> For me there is no disadvantage to support two locations for the same file and give .config the higher priority.
> 
> I would say it's more important to add a message which config file will be actually used. I'm going to add that
> 
> Thoughts ?

I would rather get rid of supporting both locations at som epoint, which
is why I suggested the warning. However, I am not bent out of shape if
we support both locations.

I could argue that having a message about which config file is being
used is just as annoying as a warning message about the file in a
"deprecated" location. On the other hand it will aid in debugging if
things go awry, thus I it is reasonable to have a "Used configuration
from: %s" %self.config_filename

-- 
rjschwei@suse.com
IRC: robjo

schaefi: Fair enough, I have changed the message to be displayed in debug mode only

bear454:
The README and manpage will need updating as well.

azurectl/account_setup.py
```diff
@@ -33,7 +33,6 @@ def __init__(self, filename):
         self.filename = filename
         try:
             if os.path.isfile(filename):
```
rjschwei: Maybe we should follow links? 

If we only allow files a shared configuration is more difficult to manage.

if os.path.isfile(filename_ or os.path.islink(filename):

I am OK with we only want to accept files, just a thought.

azurectl/config.py
```diff
@@ -32,24 +32,45 @@ class Config(object):
...
         if not account_name:
             account_name = 'default'
-        if not filename:
-            filename = self.__default_config(platform)
-        usr_config = ConfigParser.ConfigParser()
-        if not os.path.isfile(filename):
-            raise AzureAccountLoadFailed('no such config file %s' % filename)
+        if filename and not os.path.isfile(filename):
+            raise AzureAccountLoadFailed(
+                'no such config file %s' % filename
```
rjschwei: 'Could not find config file: %s' % filename

azurectl/account_setup.py
schaefi: I agree we should take that into account. I have opened another issue for it though. I'd like to prevent to mix more things into this one

schaefi:
README and manpage updated as well, thanks James for pointing that out

Are we good to go with this one now ?

PR #50, opened 2015-09-22
Added request management

schaefi:
Added support for obtaining request status and also waiting
for request completion for service management tasks

### DIFF
--- a/azurectl/azure_account.py
+++ b/azurectl/azure_account.py
@@ -57,7 +57,7 @@ class AzureAccount(object):
             return self.__get_first_subscription_id()
 
     def storage_key(self, name=None):
-        self.get_service()
+        self.__build_service_instance()
         if not name:
             name = self.storage_name()
         try:
@@ -69,7 +69,7 @@ class AzureAccount(object):
         return account_keys.storage_service_keys.primary
 
     def instance_types(self):
-        self.get_service()
+        self.__build_service_instance()
         result = []
         for rolesize in self.service.list_role_sizes():
             memory = rolesize.memory_in_mb
@@ -88,7 +88,7 @@ class AzureAccount(object):
         return result
 
     def storage_names(self):
-        self.get_service()
+        self.__build_service_instance()
         result = []
         for storage in self.service.list_storage_accounts():
             result.append(storage.service_name)
@@ -107,6 +107,10 @@ class AzureAccount(object):
         return result
 
     def get_service(self):
+        self.__build_service_instance()
+        return self.service
+
+    def __build_service_instance(self):
         if self.service:
             return
         publishsettings = self.publishsettings()
--- a/azurectl/azure_account.py
+++ b/azurectl/azure_account.py
@@ -57,7 +57,7 @@ class AzureAccount(object):
             return self.__get_first_subscription_id()
 
     def storage_key(self, name=None):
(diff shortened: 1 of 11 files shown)


schaefi:
> Is --id really a command?

no, actually a good catch we have this mix for all commands. I have separated it in the usage message now for all of them

schaefi:
> Maybe also capitalize 'log'?

don't want to do this now. the use of the global log instance is a statement of the form

```python
from logger import log
```

I agree because the log instance is application global we could change this to LOG. However I would consider this an extra refactoring task which we should not do now

ok ?

azurectl/azure_account.py
```diff
@@ -106,7 +106,7 @@ def publishsettings(self):
         )
         return result
 
-    def __get_service(self):
+    def get_service(self):
         if self.service:
             return
```
rjschwei: Based on the name of the method I would expect

return self.service

at this point??

azurectl/compute_request_task.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self.result = DataCollector()
+        self.out = DataOutput(
+            self.result,
+            self.global_args['--output-format'],
+            self.global_args['--output-style']
+        )
+
+        self.account = AzureAccount(self.config)
+        self.account.get_service()
```
rjschwei: A "get_" method should return something. Whatever it returns should then be used.

azurectl/compute_request_task.py
```diff
@@ -0,0 +1,91 @@
...
+                'request:' + format(self.command_args['--id']),
+                format(self.__get_status())
+            )
+            self.out.display()
+        elif self.command_args['wait']:
+            self.__wait_for_request_to_complete()
+
+    def __get_status(self):
+        return self.request_result.status(
+            self.account.service
```
rjschwei: Based on the name I would expect a status indicator to be returned, True, False, Pending, Busy, Completed as some status indicators that I can think of. But the method appears to return a service, which from my perspective is not a status.

rjschwei:
On 09/23/2015 06:14 AM, Marcus Schäfer wrote:
>     Maybe also capitalize 'log'?
> 
> don't want to do this now. the use of the global log instance is a
> statement of the form
> 
> from logger import log
> 
> I agree because the log instance is application global we could change
> this to LOG. However I would consider this an extra refactoring task
> which we should not do now
> 
> ok ?

It should not be LOG, all caps is intended for constants. A singleton is
not a constant.



-- 
rjschwei@suse.com
IRC: robjo

azurectl/compute_request_task.py
schaefi: ? 
return self.request_result.status(...)

it does what you wrote, it returns the status

azurectl/compute_request_task.py
schaefi: yes this is misleading I'm going to change it

azurectl/version.py
```diff
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '0.11.3'
+__VERSION__ = '0.12.3'
```
bear454: Bumping the minor version should reset the subversion: `'0.12.0'`, please.

Issue #48, opened 2015-09-18
Cannot setup accounts without having setup a [default] account.


bear454:
When setting up an account for the first time:

```
> azurectl setup account add \
> --name my-account \
> --publish-settings-file my.publishsettings \
> --storage-account-name some-storage-account \
> --container-name vhds
ERROR: AzureAccountNotFound: 'Account default not found'
```

This can also be reproduced on an existing account, by removing or renaming the `[default]` section in the config file.

schaefi:
Uhm, shame on me not having tested that. I'll fix that as soon as possible

bear454:
On Fri, 2015-09-18 at 10:59 -0700, Marcus Schäfer wrote:
> Uhm, shame on me not having tested that. I'll fix that as soon as
> possible
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/SUSE/azurectl/issues/48#issuecomment-141520422

I was going to take care of it, just wanted the PR on the file location
merged in first ;-)

-- 
James Mason
Technical Architect, Public Cloud
openSUSE Member
SUSE
jmason@suse.com
-------------------------------------
SUSECon 2015: Register at susecon.com

schaefi:
That's good, thanks so I'll leave it in your hands. Yes let's wait for the config file move to be merged

PR #52, opened 2015-09-24
Allow `setup account add` to function without an existing config file

bear454:
Addresses Issue #48


schaefi:
Hope it's ok I have rebased from master to fixup the tests and also fixed the merge conflicts with regards to version and coverage reference

bear454:
Absolutely its okay! Thanks!

azurectl/config.py
```diff
@@ -47,14 +48,14 @@ def __init__(self, account_name=None, filename=None, platform=PLATFORM):
...
+            self.config_file = self.paths.default_config()
             if not self.config_file:
                 raise AzureAccountLoadFailed(
                     'could not find default configuration file %s %s: %s' %
                     (
-                        ' or '.join(self.config_files),
+                        ' or '.join(self.paths.config_files),
                         'in home directory',
-                        self.__home_path(platform)
+                        self.paths.home_path
```
schaefi: home_path is a method, thus should be called as such

```
self.paths.hone_path()
```

azurectl/setup_account_task.py
```diff
@@ -49,18 +51,35 @@
...
 
 class SetupAccountTask(CliTask):
     """
         Process setup config commands
     """
+    def __init__(self, load_config=True):
+        """
+            Override CliTask's init, gracefully handle the case
+            where config file does not exist, so a new one may be added.
+        """
```
schaefi: Hmm, overwriting the constructor surely works but you have to call back the base class which I'd like to avoid. Couldn't we implement that as a post_init method. I have the following in mind:

in cli_task.py

```python
def __init__(self, load_config=True):
    ...
    post_init()

def post_init(self):
    pass
```

in setup_account_task.py

```python
def post_init(self):
    # do your stuff, don't care for the base class construction
```

Thoughts ?

azurectl/setup_account_task.py
```diff
@@ -49,18 +51,35 @@
...
+        try:
+            CliTask.__init__(self, load_config)
+            self.config_file = self.config.config_file
+        except AzureAccountLoadFailed:
+            t, v, tb = sys.exc_info()
```
schaefi: It would be nice if the code makes clear what t, v, tb means without having to read what sys.exc_info returns

test/unit/setup_account_task_test.py
```diff
@@ -6,6 +6,7 @@
 import nose_helper
 
 import azurectl
+from azurectl.azurectl_exceptions import *
```
schaefi: it's just the test so I don't have a strong opinion on that, but it's better to import the exceptions in question and not all of them

schaefi:
overall a great improvement, thanks

azurectl/setup_account_task.py
bear454: I tried

```
def __init__(self, load_config=True):
    ...
    if load_config:
        self.__load_config()
end

def __load_config()
    """
    override me
    """
```

But the CliTask's `__load_config` wasn't overridden by the same function in `SetupAccountTask`. Was that because of the `self` reference? Does the method need to be public to override?

azurectl/setup_account_task.py
bear454: It would be nice if there was an easier way to re-raise the same exception, but I couldn't find one.  I'll be a little more explicit.

bear454: Turns out there is an easier way... resolved in 6fb6fde .

azurectl/setup_account_task.py
bear454: Fixed in 194c29c.

test/unit/setup_account_task_test.py
bear454: I can see that in the models, but in the tests, I'd rather favor a setup that saves dev time at the expense of a few cycles. I'd like to refactor the tests to use a relatively generic loader/helper, but that was outside the scope here.

schaefi: yep ok for me

azurectl/cli_task.py
```diff
@@ -26,7 +26,7 @@ class CliTask(object):
         the interface to the command options and the account to use
         for the task
     """
-    def __init__(self, load_config=True):
+    def __init__(self, should_load_config=True):
```
schaefi: That one broke azurectl/app.py, please update the instance creation in app.py as follows

```
app = CliTask(should_load_config=False)
```

azurectl/config_file_path.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class ConfigFilePath(object):
+    """
+        File path management for config files
+    """
+    PLATFORM = sys.platform[:3]
+
+    def __init__(self, filename=None, platform=PLATFORM):
+        from logger import log
```
schaefi: 'log' imported but unused

azurectl/config_file_path.py
```diff
@@ -0,0 +1,60 @@
...
+import sys
+
+
+class ConfigFilePath(object):
+    """
+        File path management for config files
+    """
+    PLATFORM = sys.platform[:3]
+
+    def __init__(self, filename=None, platform=PLATFORM):
```
schaefi: Unused argument 'filename'

azurectl/setup_account_task.py
```diff
@@ -42,6 +42,8 @@
         subscription id, if more than one subscription is included in your
         publish settings file.
 """
+import sys
```
schaefi: Unused import sys

azurectl/cli_task.py
bear454: Duh. Thanks. Fixed in d799251.

azurectl/config_file_path.py
bear454: Fixed in 3a6563d.

azurectl/config_file_path.py
bear454: Fixed in 3a6563d.

azurectl/setup_account_task.py
bear454: Fixed in 3a6563d.

schaefi:
all good now, thanks much

Issue #51, opened 2015-09-24
'default' is not a valid name when adding an account to config...


bear454:
Found this while working on #48 : 
https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.add_section

schaefi:
That basically means you can have only one default section. Is this a problem ?

If I call ./azurectl setup account add --name default ... I got

```
ERROR: AzureConfigAddAccountSectionError: 'ValueError: Invalid section name: default'
```

If I use another section name which also exists I got

```
ERROR: AzureConfigAddAccountSectionError: "DuplicateSectionError: Section 'sharetest' already exists"
```

So I think we are good if catch the ValueError on default and provide a better error message

Thoughts ?

bear454:
Please explain how you would write the section named 'default' into the config file, for example, in the case of a new user.

schaefi:
```
[default]
storage_account_name =
storage_container_name =
publishsettings =

[james]
storage_account_name =
storage_container_name =
publishsettings =

[marcus]
storage_account_name =
storage_container_name =
publishsettings =
```

I'm sure I'm missing the point ?

bear454:
The point is, you cannot produce that '[default]' section using `azurectl setup account --name=default ...`.

schaefi:
ok got it, will look into this today

PR #53, opened 2015-09-28
Config parser default section handling

schaefi:
The DEFAULT section in an INI file is handled special by python's config parser implementation. Therefore this patch refactors the way we use the config file sections. Any section of the config file belongs to an account, the DEFAULT section allows to select one of those accounts as the default account with the default_account keyword.

In addition the  

```
azurectl setup account default --name
```

subcommand to handle the default account was added.
This fixes Issue #51

### DIFF
--- a/azurectl/account_setup.py
+++ b/azurectl/account_setup.py
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import ConfigParser
+from ConfigParser import ConfigParser
 import os
 
 # project
@@ -29,7 +29,7 @@ class AccountSetup(object):
         Implements setup methods to list, add and delete account configs
     """
     def __init__(self, filename):
-        self.config = ConfigParser.ConfigParser()
+        self.config = ConfigParser()
         self.filename = filename
         try:
             if os.path.isfile(filename):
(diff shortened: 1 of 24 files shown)


azurectl/account_setup.py
```diff
@@ -45,16 +45,22 @@ def list(self):
...
         return accounts
 
     def remove(self, name):
         """
             remove specified account section
         """
+        if name == self.get_default_account():
+            log.info('Section %s is the default section', name)
+            log.info('Please setup a new default prior to removing')
+            return False
```
rjschwei: Not certain I follow the logic here:
The user cannot remove [default] that's fair enough. However it is not possible to designate, lets say [myacount] as default as the default is only indicated by the section declaration. Therefore it is not possible to follow the advise of the message. By modifying [default], manually I presume, I explicitly eliminate the need to remove it. Wouldn't it make more sense to just tell the user that the [default] account cannot be removed?

azurectl/account_setup.py
```diff
@@ -94,6 +100,20 @@ def add(
...
+        return True
+
+    def get_default_account(self):
+        return self.config.defaults()['default_account']
+
+    def set_default_account(self, section_name):
+        if section_name not in self.config.sections():
+            log.info('Section %s does not exist', section_name)
+            return False
+        self.config.defaults()['default_account'] = section_name
```
rjschwei: OK, so we allow lets say [myaccount] to be designated as default account. How is this represented in the file such that the default remains persistent across invocations?

azurectl/config.py
```diff
@@ -21,7 +21,8 @@
     AzureAccountLoadFailed,
     AzureConfigParseError,
     AzureAccountNotFound,
-    AzureAccountValueNotFound
+    AzureAccountValueNotFound,
+    AzureAccountDefaultSectionNotFound
```
rjschwei: Order should be reversed to stay in alpha order

azurectl/setup_account_task.py
```diff
@@ -114,6 +119,15 @@ def __add(self):
...
 
+    def __set_default(self):
+        if self.setup.set_default_account(
+            self.command_args['--name']
+        ):
+            log.info(
+                'Account %s is now set as default account',
+                self.command_args['--name']
+            )
+
```
rjschwei: It appears to me that somehow we would need to persist the information to disk??

Issue #54, opened 2015-10-01
certificate creation is (now) async; interferes with VM creation


bear454:
Recently, VM creation started to fail, with the following error:

```
$ azurectl compute vm create --image-name="b4590d9e3ed742e4a1d46e5424aa335e__suse-sles-11-sp4-v20150714" --cloud-service-name="..." --region="West US" --instance-type=Medium --ssh-private-key=... --user=...

ERROR: AzureVmCreateError: 'AzureHttpError: Bad Request\n<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>BadRequest</Code><Message>The certificate with thumbprint ... was not found.</Message></Error>'
```

Further digging shows that uploading the ssh key to an Azure management certificate is asynchronous; I suspect the issue is due to the create command not waiting for the certificate to be created.

PR #55, opened 2015-10-01
When creating a VM using an ssh key, wait for service to create the key dependency before continuing on.

bear454:
Resolves issue #54


schaefi:
ok seems my patch send by mail worked to fix this issue. Thanks for testing and merging

bear454:
It basically fixed the cert issue - I needed to add a test change as well - but that opened a whole new can of worms: #57 .

PR #56, opened 2015-10-01
Add a script that enables running azurectl commands from the git project dir.

bear454:

PR #58, opened 2015-10-02
Revert "Add a script that enables running azurectl commands from the …

schaefi:
Reverts SUSE/azurectl#56

Such a script already exists in bin/azurectl


schaefi:
DId not understand why we need an additional one in the main tree

```bash
alias azurectl='cd ~ms/Project/azurectl/bin && ./azurectl'

azurectl ...
```

I'm a friend of less code to maintain :)

bear454:
That's fine; I'll maintain my dev tool in my `~/bin`; just trying to share something I found useful.

Issue #57, opened 2015-10-01
cannot create a vm outside of the region a storage account is in


bear454:
Given an azurectl account with a storage account in region X, VMs cannot be created in any region except X.

Trying to instantiate a VM in a different region throws the following error (the storage account used here was in West US, the argument was `--region="East US 2"`):

```
ERROR: AzureVmCreateError: 'AzureHttpError: Bad Request\n<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>BadRequest</Code><Message>The location or affinity group West US of the storage account where the source image b4590d9e3ed742e4a1d46e5424aa335e__suse-sles-11-sp4-v20150714 resides is not in the same location or affinity group as the specified cloud service. The source image must reside in a storage account that has the same affinity group or location as the cloud service East US 2.</Message></Error>'
```

*Note:* This error is currently masked by another error:

```
ERROR: AzureVmCreateError: 'AzureHttpError: Bad Request\n<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>BadRequest</Code><Message>MediaLocation is only applicable when using a published VM Image. It cannot be used with user VM Images, OS or data virtual hard disks.</Message></Error>'
```

You must patch out `virtual_machine.py` line 120 to work around this error.

I'm including both here, because there's obviously some issues around location handling that need to be solved in both VM creation and the config file; it would be better to resolve the overall issue than to just patch around the the 2nd error.

rjschwei:
We need to split the region dependent information from the account information. My suggestion is to follow the pattern of the config file we use for the ec2utils.

[account-]
[region-]

In the region section we can store the storage account information as follows

[region-us-east]
default_storage=mybucket
storage_accounts=a_bucket,b_bucket,c_bucket

The [account-] section holds information about Azure access.

If a user instantiates and instance with --region we look in the config file for a storage account. If there is none configured for the given region we create one and add the information to the config file. If a storage account exists we use the value for default_storage unless otherwise directed by a --storage-account command line option.

schaefi:
ok I like the idea to have region sections containing storage and container information

I don't like to automatically create a storage account, this sounds like making possibly unwanted user decisions. Overall I don't like tools which tries to be more intelligent than the user :) Instead if no config for the region exists an error should be thrown

[...]

PR #59, opened 2015-10-06
Account / Region separation

schaefi:
Change configuration file structure
    
Instead of one account section containing information about the Azure account as well as information about the storage/container account it is required to handle these information in separate sections. The reason for this change is that storage accounts can exist in different regions. Therefore the storage accounts should be placed in a region section. This refactoring also impacts commands with --region and/or --container  options because they are now handled as global options.
    
The new configuration format is as follows:

```    
[DEFAULT]
default_account = account:name
default_region  = region:name
    
[account:some-account]
publishsettings = filename
    
[region:some-region]
default_storage_account   = name
default_storage_container = name
storage_accounts   = comma-list
storage_containers = comma-list
```

The account setup command allows to perform the following actions
    
* add account and region sections
* remove sections
* set default account
* set default region
    
There is still room for improvement on the account setup command. However that should be part of an additional implementation. This fixes Issue #57

### DIFF
--- a/azurectl/version.py
+++ b/azurectl/version.py
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '0.12.1'
+__VERSION__ = '1.0.0'
--- a/doc/man/azurectl.md
+++ b/doc/man/azurectl.md
@@ -52,11 +52,6 @@ Azure image management
 
 # GLOBAL OPTIONS
 
-## __--debug__
-
-Enable debugging mode. In this mode azurectl is more verbose and
-provides information useful to clarify processing issues.
-
 ## __--config=file__
 
 Location of global config file, default is searched in:
@@ -64,9 +59,23 @@ Location of global config file, default is searched in:
 1. __~/.config/azurectl/config__
 2. __~/.azurectl/config__
 
+The azurectl config file is an INI style file structured into sections. There are account and region sections which are handled by the __azurectl setup account command__
+
 ## __--account=name__
 
-Account name to select from config file. The azurectl config file is an INI style file structured into sections. Each section takes a name and is selectable via the --account option. The default account section name if no account is selected is: __default__
+Account name to use for operations. By default the account referenced as __default_account__ from the the DEFAULT section will be used. In the configuration file the account section is stored with a prefix named __account:<value>__. The given value must match one of the account sections.
+
+## __--region=region__
+
+Region name to use for operations. By default the region referenced as __default_region__ from the DEFAULT section will be used. An azure region is the name of the geographic region to run e.g virtual instances or store VHD disk images. An example region name would be __East US 2__. The value for the region has to match the region names used in Microsoft Azure. In the configuration file the region section is stored with a prefix named __region:<value>__. The given value must match one of the region sections.
+
+## __--storage-account=name__
+
+Storage account name to use for operations. The account name must be part of the __storage_accounts__ setup of the associated region in the configuration file
+
+## __--storage-container=name__
+
+Storage container name to use for operations. The container name must be part of the __storage_containers__ setup of the associated region in the configuration file
 
 ## __--output-format=format__
 
@@ -76,7 +85,7 @@ Print information in specified format. Supported formats are
 
 The default format is: json
 
-## __--output-style=<style>__
+## __--output-style=style__
 
 Print information in specified style. Supported styles are
 
@@ -84,3 +93,10 @@ Print information in specified style. Supported styles are
 * standard
 
 The default style is: standard
+
+## __--debug__
+
+Enable debugging mode. In this mode azurectl is more verbose and
+provides information useful to clarify processing issues.
+
+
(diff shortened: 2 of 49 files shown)


schaefi:
along with the change you can now globally select account, region, storage account in region and container in region. It's still possible to select a wrong storage account vs. container name combination. However this will lead to an expected error message saying this container does not exist in this storage account. Thus this is ok for me so far

If we also want to disallow invalid storage vs. container combination we need another config file refactoring.

schaefi:
So far this is a big change but I did not see how to make it in smaller junks. Hope it's ok for a review

azurectl/version.py
```diff
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '0.12.1'
+__VERSION__ = '0.13.0'
```
rjschwei: According to semantic version we should bump to 1.0.0 as this contains a number of incompatible changes. At the command line level the --region argument is now provided in a different place than previously. At the SPI level the constructor for config.py is incompatible with previous implementations.

I realize that 1.0.0 also has other meanings, as in "first stable release". Well we are reasonably close and I'd really like us to be consistent with our version indicators. Thus, IMHO it should be 1.0.0

doc/man/azurectl.md
```diff
@@ -64,9 +59,23 @@ Location of global config file, default is searched in:
 1. __~/.config/azurectl/config__
 2. __~/.azurectl/config__
 
+The azurectl config file is an INI style file structured into sections. There are account and region sections which are handled by the __azurectl setup account command__
+
```
rjschwei: Suggested addition:

"""
While region sections [REGIONNAME] do not have to match the region names used in Mivrosoft Azure it is recommended to use the Microsoft Azure region names to avoid confusion. 
"""

doc/man/azurectl.md
```diff
@@ -64,9 +59,23 @@ Location of global config file, default is searched in:
...
 ## __--account=name__
 
-Account name to select from config file. The azurectl config file is an INI style file structured into sections. Each section takes a name and is selectable via the --account option. The default account section name if no account is selected is: __default__
+Account name to use for operations. By default the account referenced as __default_account__ from the the DEFAULT section will be used
+
```
rjschwei: Suggested addition:

"""
The value of the option must match one of the section names in the configuration file being processed.
"""

doc/man/azurectl.md
```diff
@@ -64,9 +59,23 @@ Location of global config file, default is searched in:
...
+## __--region=region__
+
+Region name to use for operations. By default the region referenced as __default_region__ from the DEFAULT section will be used. An azure region is the name of the geographic region to run e.g virtual images or store VHD disk image. An example region name could be __East US 2__
+
```
rjschwei: Suggested addition:

"""
While value does not have to match the region names used in Microsoft Azure it is recommended to use the Microsoft Azure region names to avoid confusion. The value of the option must match one of the section names in the configuration file being processed.
"""

doc/man/azurectl::setup::account.md
```diff
@@ -4,11 +4,8 @@ azurectl - Command Line Interface to manage Microsoft Azure
 
 # SYNOPSIS
 
-__azurectl__ setup account add --name=*configname*
+__azurectl__ setup account add --name=*account_name* --publish-settings-file=*file*
```
rjschwei: In the change fro the Readme.md file it looks as if "add" was renamed to "configure" this could be a false impression because of missing context in the diff.

rjschwei:
Overall it looks good. Personally I would have preferred to different the config section names with [account-] and [region-] prefixes as it provides a visually more explicit differentiation between the 2 section types when viewing the config file in an editor. Adding the name prefix would require some "magic" in the code, however, I think the visual clarity in the file would be well worth the trade off.

However, I am OK with the non explicit section separation as currently implemented. With the current implementation a section named [foo] could be either an account or a region, one has to look at the options of the section to determine the section type.

bear454:
This is a pretty big change, @schaefi , but I agree there's no good way around that. The code & test coverage look good; lets get the docs and the version number fixed up, and merge this in!

doc/man/azurectl::setup::account.md
schaefi: with the configure command you create a complete configuration (account, region, storage) previously we had all that combined in one section and just provide to add one or more of them.

The add command has changed to be one of

* account add
* account region add

So we did not rename the add command but extended the setup with a configure command which I find useful to start with azurectl when there is no configuration file available. This was also the reason why I put it into the README

schaefi:
I will add account and region prefixes to the configuration file. I'm not going to expose them to the user though. They will be handled internally and you can see them in the config file but from a user perspective you say --region "East US 2" and not --region "region:East US 2"

doc/man/azurectl.md
schaefi: the value for the region name must match an azure region because we use this information when running an instance

doc/man/azurectl.md
schaefi: The region name must match one of the azure regions, this information is passed e.g when we run an instance

bear454: If the region *must match* one of the Azure regions, how does one create a config file with multiple accounts? Storage accounts in azure are name-unique, so you cannot use the same region data for two accounts; e.g. azure-rd and azure-updates.

Please explain how I can use the config file, in this way, to access both accounts' storage?

Issue #61, opened 2015-10-15
Delete pointless config file options


schaefi:
Config options:

```
storage_accounts
storage_containers
```

This information is used to check if a given container or storage name given on the command line is part of the list. However that is no real validation because the user could have added wrong information in the config file as well. Thus the only truth is the result from the api call

There is no need to run a validation check for this values given on the command line. Information from the command line is just used as passed and will eventually create an error from the api call

schaefi:
done now

Issue #60, opened 2015-10-15
Change meaning of --account option


schaefi:
In an effort to maintain multiple configuration files the --account option should not select an account section in the config file but should select a config file from the default directory. Example

```
azurectl --account suse-rd
```

looks up

```
~/.config/azurectl/config.suse-rd
```

the loaded config file should only contain one account section which is referenced in the DEFAULT section.

bear454:
IMO `suse-rd.config` is more conventional, as [description].[type] .

schaefi:
yes makes sense, thanks. working on this now

schaefi:
done now

PR #63, opened 2015-10-16
Code refactoring Issues #60 and #61

schaefi:
Refactor account handling
    
Instead of managing multiple account sections in one configuration file we allow to select a configuration file by a template name The former --account parameter specified a section name in a configuration file. With this patch --account specifies a template name for a configuration file with the name
    
* ~/.config/azurectl/<template_name>.config
    
Along with this change the options --config and --account are  mutually exclusive. --config still allows to specify any filename. The basic structure of the configuration has not changed. Thus it is still possible to maintain multiple account sections in the configuration file, however only one can be actively selected  by the default_account attribute from the DEFAULT section.
    
Switching between accounts can be achieved by selecting a new default account via 'azurectl setup account default --name ...'  but no longer via the --account option.

In addition the configuration attributes

* storage_accounts
* storage_containers

were removed

### DIFF
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -149,10 +149,6 @@ class AzureAccountLoadFailed(AzureError):
     pass
 
 
-class AzureStorageAccountInvalid(AzureError):
-    pass
-
-
 class AzureStorageFileNotFound(AzureError):
     pass
 
--- a/azurectl/cli.py
+++ b/azurectl/cli.py
@@ -18,8 +18,7 @@ usage: azurectl -h | --help
                 [--output-style=<style>]
                 [--debug]
            setup <command> [<args>...]
-       azurectl [--config=<file>]
-                [--account=<name>]
+       azurectl [--config=<file> | --account=<name>]
                 [--region=<name>]
                 [--storage-account=<name>]
                 [--storage-container=<name>]
@@ -32,8 +31,8 @@ usage: azurectl -h | --help
 
 global options:
     --account=<name>
-        account name in config file, default is default_account
-        from config file DEFAULT section
+        account template name. The given name is used as config file
+        template: ~/.config/azurectl/<name>.config.
     --config=<file>
         config file, default is: ~/.config/azurectl/config
     --output-format=<format>
--- a/azurectl/cli.py
+++ b/azurectl/cli.py
@@ -43,13 +43,13 @@ global options:
         region name in config file, default is default_region
         from config file DEFAULT section
     --storage-account=<name>
-        storage account name in config file. The account name must be
-        part of the storage_accounts setup of the associated region
-        in the configuration file
+        storage account name to use for operations. This will
+        take precedence over the configured default_storage_account
+        from the config file
     --storage-container=<name>
-        storage container name in the config file. The container name
-        must be part of the storage_containers setup of the associated
-        region in the configuration file
+        storage container name to use for operations. This will
+        take precedence over the configured default_storage_container
+        from the config file
     -v --version
         show program version
     help
--- a/azurectl/cli.py
+++ b/azurectl/cli.py
@@ -31,8 +31,9 @@ usage: azurectl -h | --help
 
 global options:
     --account=<name>
-        account template name. The given name is used as config file
-        template: ~/.config/azurectl/<name>.config.
+        account name. The given name value is used to select the configuration
+        file of the form <name>.config from the configuration location
+        ~/.config/azurectl.
     --config=<file>
         config file, default is: ~/.config/azurectl/config
     --output-format=<format>
--- a/azurectl/config.py
+++ b/azurectl/config.py
@@ -39,38 +39,24 @@ class Config(object):
     PLATFORM = sys.platform[:3]
 
     def __init__(
-        self, account_name=None, region_name=None,
-        storage_account_name=None, storage_container_name=None,
-        filename=None, platform=PLATFORM
+        self,
+        account_file_template_name=None,
+        region_name=None,
+        storage_account_name=None,
+        storage_container_name=None,
+        filename=None,
+        platform=PLATFORM
     ):
         from logger import log
 
-        paths = ConfigFilePath(platform)
-
-        self.account_name = account_name
-        self.region_name = region_name
         self.storage_container_name = storage_container_name
         self.storage_account_name = storage_account_name
 
-        self.config = ConfigParser()
+        self.config_file = self.__lookup_config_file(
+            platform, account_file_template_name, filename
+        )
 
-        if filename and not os.path.isfile(filename):
-            raise AzureAccountLoadFailed(
-                'Could not find config file: %s' % filename
-            )
-        elif filename:
-            self.config_file = filename
-        else:
-            self.config_file = paths.default_config()
-            if not self.config_file:
-                raise AzureAccountLoadFailed(
-                    'could not find default configuration file %s %s: %s' %
-                    (
-                        ' or '.join(paths.config_files),
-                        'in home directory',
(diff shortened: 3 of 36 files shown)


azurectl/cli.py
```diff
@@ -32,8 +31,8 @@
 
 global options:
     --account=<name>
-        account name in config file, default is default_account
-        from config file DEFAULT section
+        account template name. The given name is used as config file
+        template: ~/.config/azurectl/<name>.config.
```
rjschwei: I would not use the term "template". In general people associate something different with a template than what I think we are after. How about this:

"""
account name. The given name value is used to select the configuration file of the form <name>.config from the configuration location ~/.config/azurectl.
"""

azurectl/azurectl_exceptions.py
```diff
@@ -149,10 +149,6 @@ class AzureAccountLoadFailed(AzureError):
     pass
 
 
-class AzureStorageAccountInvalid(AzureError):
-    pass
-
-
```
rjschwei: I am not certain I understand why this exception is no longer needed. It appears to me that if the user specifies an invalid storage account with the --storage-account option this error condition would still apply?

azurectl/config.py
```diff
@@ -39,38 +38,24 @@ class Config(object):
     PLATFORM = sys.platform[:3]
 
     def __init__(
-        self, account_name=None, region_name=None,
-        storage_account_name=None, storage_container_name=None,
-        filename=None, platform=PLATFORM
+        self,
+        account_file_template_name=None,
```
rjschwei: I think the variable should be named account_filename or account_file_name as it is not a template that we will be customizing. The use of "template" in this context is confusing to me.

azurectl/config.py
```diff
@@ -80,25 +65,16 @@ def __init__(
...
 
-        defaults = self.config.defaults()
-
-        if not defaults:
+        if not self.config.defaults():
             raise AzureAccountDefaultSectionNotFound(
-                'could not find default section in configuration file %s' %
+                'Empty or undefined default section in configuration file %s' %
                 self.config_file
             )
```
rjschwei: We should not throw this exception if the user provides all the information necessary on the command line. We should only complain about a missing default section if we actually need to access the information that would be set in the default section.

The current check is not sufficient for this condition.

azurectl/config.py
```diff
@@ -176,3 +130,63 @@ def __get_region_option(self, option):
                 (option, self.region_name, self.config_file)
             )
         return result
+
+    def __lookup_config_file(self, platform, template, filename):
+        paths = ConfigFilePath(template, platform)
```
rjschwei: The use of "template" as a name is confusing IMHO

azurectl/azurectl_exceptions.py
schaefi: the account meaning has changed, formerly it was a section name which could be invalid, now it's a file path and the error you get is an AzureAccountLoadFailed with an appropriate message: 'Could not find account config file: %s %s: %s'

azurectl/config.py
schaefi: but this is not a file name, we use this name and construct a file name out of it, so it's partial information for a file name which I consider a template. maybe the term template is not fulfilled here but it's also not a file name and the reason why I did not name it as such

azurectl/config.py
schaefi: the information from the default section is accessed directly one line below this check which is the reason why this check makes sense. This section is mandatory to have any default and read in with constructing a Config instance.

azurectl/config.py
schaefi: maybe this is only a language problem but for me template means not complete, customizable, extensible. I have not so many objections but I'm open to change this if it makes sense, maybe 

* config_file_prefix
* config_file_name_prefix

does that sound any better ?

azurectl/config.py
rjschwei: On 10/19/2015 11:19 AM, Marcus Schäfer wrote:
>> @@ -39,38 +38,24 @@ class Config(object):
>>      PLATFORM = sys.platform[:3]
>>  
>>      def __init__(
>> -        self, account_name=None, region_name=None,
>> -        storage_account_name=None, storage_container_name=None,
>> -        filename=None, platform=PLATFORM
>> +        self,
>> +        account_file_template_name=None,
> 
> but this is not a file name, we use this name and construct a file name out of it, so it's partial information for a file name which I consider a template. maybe the term template is not fulfilled here but it's also not a file name and the reason why I did not name it as such

If the file name is not constructed yet then just keep it as
account_name which is abstract in that it could be used to construct a
filename later or a section name. Thus using the "account_name" variable
name does not imply any specific action, while using "template" does
imply certain conditions.
> 


-- 
rjschwei@suse.com
IRC: robjo

azurectl/config.py
rjschwei: On 10/19/2015 11:24 AM, Marcus Schäfer wrote:
>>              raise AzureAccountDefaultSectionNotFound(
>> -                'could not find default section in configuration file %s' %
>> +                'Empty or undefined default section in configuration file %s' %
>>                  self.config_file
>>              )
> 
> the information from the default section is accessed directly one line below this check which is the reason why this check makes sense. This section is mandatory to have any default and read in with constructing a Config instance.

Fair enough, is it documented that the default section must exist?

-- 
rjschwei@suse.com
IRC: robjo

azurectl/config.py
rjschwei: On 10/19/2015 11:26 AM, Marcus Schäfer wrote:
>> @@ -176,3 +130,63 @@ def __get_region_option(self, option):
>>                  (option, self.region_name, self.config_file)
>>              )
>>          return result
>> +
>> +    def __lookup_config_file(self, platform, template, filename):
>> +        paths = ConfigFilePath(template, platform)
> 
> maybe this is only a language problem but for me template means not complete, customizable, extensible. I have not so many objections but I'm open to change this if it makes sense, maybe 
> 
> * config_file_prefix
> * config_file_name_prefix
> 
> does that sound any better ?

just call it account_name. It is descriptive about what is being
expected, a name vs a "template" that may be customized. What the name
is used for in the end is at this point somewhat immaterial. An argument
named "template" couls also be this:

'[account-%s]\nname=bob\nlocation=%s\n'

That's a template and if I look at "template" as the argument  name then
I would expect to pass something like that or a template object.



-- 
rjschwei@suse.com
IRC: robjo

azurectl/config.py
schaefi: now it is

schaefi:
ok I think we are good to go

bear454:
On Mon, 2015-10-19 at 13:29 -0700, Marcus Schäfer wrote:
> ok I think we are good to go
> —
> Reply to this email directly or view it on GitHub.

I'm still not happy with the language at:
https://github.com/SUSE/azurectl/commit/299c30a34267aa995588ac4094f11ae
7b37b1869#diff-9fe92f96046100807b249f8eebf790d7R34

Would you prefer I add a patch for that after?
-- 
James Mason
Technical Architect, Public Cloud
openSUSE Member
SUSE
jmason@suse.com
-------------------------------------
SUSECon 2015: Register at susecon.com

schaefi:
>    I'm still not happy with the language at:
>    https://github.com/SUSE/azurectl/commit/299c30a34267aa995588ac4094f11ae
>    7b37b1869#diff-9fe92f96046100807b249f8eebf790d7R34

Sorry don't see it, I have eliminated the use of 'template' in
commit 7ca095410ae9446c336cb9026e2725066187fe11. Did I forgot something ?

>    Would you prefer I add a patch for that after?

patch it, I'm happy to see patches :)

will go to bed now

Thanks

Issue #62, opened 2015-10-15
VM creation needs to be aware of media_link image attribute


bear454:
Some Azure images, especially self-uploaded images, provide a `media_link`. Example:

```
    {
      "affinity_group": "",
      "category": "User",
      "description": "SLES12-JeOS-for-MS-HyperV",
      "label": "SLES12-JeOS-for-MS-HyperV",
      "location": "East US",
      "media_link": "https://suseazurerd.blob.core.windows.net/vm-images/SLES12-JeOS-for-MS-HyperV.x86_64-1.13.1.vhd",
      "name": "SLES12-JeOS-for-MS-HyperV",
      "os": "Linux"
    }
```

When calling `ServiceManagementService#create_virtual_machine_deployment`, the `instance_record` needs to include the `media_location` in this case, which our code handles properly.

----

Some Azure images, in particular, published images, *do not* provide a `media_link`. Example:

```
    {
      "affinity_group": "",
      "category": "Public",
      "description": "Confidently run your production workloads on SUSE Linux Enterprise Server on Microsoft Azure, knowing that your service levels are guaranteed, and that help from SUSE and Microsoft engineers, if you need it, is just a phone call away.  SUSE Linux Enterprise Server is the only supported Enterprise Linux distribution available on Microsoft Azure, and the only one supported with tools like SUSE Studio ( [https://susestudio.com|https://susestudio.com] ) for quick and easy image creation, and SUSE Manager ( [https://www.suse.com/products/suse-manager/|https://www.suse.com/products/suse-manager/] ) for monitoring and managing your Linux virtual machines using Microsoft Systems Center.",
      "label": "SUSE Linux Enterprise Server 12",
      "location": "East Asia;Southeast Asia;Australia East;Australia Southeast;Brazil South;North Europe;West Europe;Central India;South India;West India;Japan East;Japan West;Central US;East US;East US 2;North Central US;South Central US;West US",
      "media_link": "",
      "name": "b4590d9e3ed742e4a1d46e5424aa335e__suse-sles-12-v20150512",
      "os": "Linux"
    },
```

When calling `ServiceManagementService#create_virtual_machine_deployment`, the `instance_record` *should not*  include the `media_location` in this case, which our code *does not* handle properly, causing the following error:

```
ERROR: AzureVmCreateError: 
'AzureHttpError: Bad Request

<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
[...]

PR #70, opened 2015-10-19
Don't provide the media_link attribute when creating a VM

bear454:
Resolves https://github.com/SUSE/azurectl/issues/62 .


schaefi:
Looks all good, thanks much

Issue #67, opened 2015-10-19
pre check for cloudservice DNS name already in use


schaefi:
when creating a cloudservice Azure also creates a DNS name to address it. if the DNS is already taken a message somewhat confusing to the user is displayed.

We should add a check prior to creating the cloud service and see if the DNS name Azure would use is already in use.

For building up the DNS query a lookup table taking the region name into account should be used.
e.g

```python
azure_domain{
    'West US': 'cloudapp.net'
}
```

schaefi:
There are following regions:

* East Asia
* Southeast Asia
* Australia East
* Australia Southeast
* Brazil South
* North Europe
* West Europe
* Central India
* South India
* West India
* Japan East
* Japan West
* Central US
* East US
* East US 2
* North Central US
* South Central US
* West US

schaefi:
working on this one now

schaefi:
done

PR #71, opened 2015-10-20
Added check for cloudservice URL

schaefi:
A requested cloud service will be registered as public address in Azure If this address is already in use the service comes back with a message that is not clear to the user how to fix the situation. This patch adds a DNS check prior to creating a new cloud service and provides a better error message to the user. This fixes Issue #67

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,8 @@ config = {
         'pyliblzma==0.5.3',
         'azure_storage==0.20.0',
         'azure_servicemanagement_legacy==0.20.0',
-        'python-dateutil==2.4'
+        'python-dateutil==2.4',
+        'dnspython==1.12.0'
     ],
     'packages': ['azurectl'],
     'entry_points': {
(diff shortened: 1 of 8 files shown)


schaefi:
yes I had the same feelings and just started that way. If we don't use a specific DNS as we do right now the default is the host system resolv.conf setup. dnspython will use that. However I was not sure if this is the right way to go because you don't know about its configuration state and thus the result is basically undefined

It's a matter of trust

let's see what Robert says

schaefi:
I updated the message text, thanks

schaefi:
ok done, we now use the host DNS setup and I updated the text

setup.py
```diff
@@ -18,7 +18,8 @@
         'pyliblzma==0.5.3',
         'azure_storage==0.20.0',
         'azure_servicemanagement_legacy==0.20.0',
-        'python-dateutil==2.4'
+        'python-dateutil==2.4',
+        'dnspython==1.12.0'
     ],
```
rjschwei: I don't think we need to be so strict with the dependencies for most, spare the azure-python-sdk, a > than condition should suffice.

However that can be changes in another go around.

Issue #69, opened 2015-10-19
instance check: region for image and cloud service must be the same


schaefi:
The location of the image must be the same region as the cloud service

schaefi:
working on this Issue

schaefi:
done

Issue #68, opened 2015-10-19
instance check: region for cloud service and storage account must be the same


schaefi:
Before running an instance in an already existing cloud service we have to check if the region the cloud service runs in is the same as the region the storage account exists in

schaefi:
Working on this Issue

schaefi:
done

PR #72, opened 2015-10-20
Added check for region consistency

schaefi:
When running an instance of an image the following region constraints must be fulfilled

* storage account region and cloud service region matches
* image exists in cloud service region

This patch adds those checks prior to running an instance and provides useful error messages to the user. This fixes Issue #69 and Issue #68

### DIFF
--- a/azurectl/virtual_machine.py
+++ b/azurectl/virtual_machine.py
@@ -106,15 +106,12 @@ class VirtualMachine(object):
         """
         if not self.__storage_reachable_by_cloud_service(cloud_service_name):
             message = [
-                'Storage account "%s" is not in the cloud service region "%s".',
-                'Please select a storage account matching the cloud',
-                'service region, or create a cloud service in the storage',
-                'account region "%s"'
+                'The cloud service "%s" and the storage account "%s"',
+                'are not in the same region, cannot launch an instance.'
             ]
             raise AzureStorageNotReachableByCloudServiceError(
                 ' '.join(message) % (
-                    self.account_name, self.service_location,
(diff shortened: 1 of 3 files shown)


schaefi:
I always struggle with good error messages :) feel free to rephrase them

azurectl/virtual_machine.py
```diff
@@ -102,6 +104,34 @@ def create_instance(
         """
             create a virtual disk image instance
         """
+        if not self.__storage_reachable_by_cloud_service(cloud_service_name):
+            message = [
+                'Storage account "%s" is not in the cloud service region "%s".',
+                'Please select a storage account matching the cloud',
+                'service region, or create a cloud service in the storage',
+                'account region "%s"'
```
rjschwei: I think we can be less verbose.

'The cloud service "%s" and the storage account "%s" are not in the same region, cannot launch an instance.'

azurectl/virtual_machine.py
```diff
@@ -102,6 +104,34 @@ def create_instance(
...
+                )
+            )
+
+        if not self.__image_reachable_by_cloud_service(
+            cloud_service_name, disk_name
+        ):
+            message = [
+                'Image "%s" does not exist in the region "%s"',
+                'of the running cloud service "%s".',
+                'Please select an image available to the cloud service.'
```
rjschwei: 'The selected image "%s" is not available in the region of the selected cloud service "%s", cannot launch instance'

schaefi:
Is this one now also ready to be merged ?

PR #74, opened 2015-11-04
Adaptions to replication/publishing SDK patches

schaefi:
My patches to the Azure SDK were accepted, however I had to rename the methods and the class where they live after review by $MS. Thus these changes have an effect on azurectl too

### DIFF
--- a/azurectl/image.py
+++ b/azurectl/image.py
@@ -13,7 +13,6 @@
 #
 from tempfile import NamedTemporaryFile
 from azure.servicemanagement import ServiceManagementService
-from azure.servicemanagement import ComputeManagementService
 from azure.storage.blob import BlobService
 
 # project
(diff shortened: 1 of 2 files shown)


schaefi:
Please no merge until the following PR was accepted by $MS

https://github.com/Azure/azure-sdk-for-python/pull/497

bear454:
@schaefi , do we need to wait for the API pull request before merging this in?

schaefi:
no we don't have to wait, we can merge this in as soon as we are ok with it

However the package release is kind of blocked until $MS has merged my PR. The next package release should also care for a version dependency on the sdk. That's because we have released the sdk with my patches but $MS wants my code to live in a another namespace which is an incompatible
change.

I'll care for this when I submit new packages

schaefi:
yay patches merged to the microsoft azure sdk :)

schaefi:
ok to merge this ? I need a thumbs up :)

bear454:
Yes, merge it!

James Mason
Technical Architect, Public Cloud
openSUSE Member
SUSE
jmason@suse.com

Issue #73, opened 2015-10-21
review dependencies


rjschwei:
Review the dependencies in setup.py. I think for most, spare the azure-python-sdk we can live with a > dependency rather than requiring the exact version, dns module is an example.

schaefi:
it would be only relevant for people installing azurectl via 'pyton setup.py build & install' or from pip if we want to make it available there too.

In principal I agree, however you don't know exactly if the program works with another version of the modules.

schaefi:
do we really want to extend the setup.py module requirements to version we don't know if they work ?

That's basically my question and I'm ok if a majority of the team has no problem with it

rjschwei:
Python does not have a "multi-version" solution. Thus we need to be a bit more relaxed about the dependencies, IMHO. I agree that there is risk in opening up the dependency, on the other hand I hate to be so paranoid.

Lets get James' thoughts on this.

bear454:
It seems the python community has an answer for everything, so I went in search for it.

I believe what we want is `~=` instead of `==`.

See [pep440](https://www.python.org/dev/peps/pep-0440/#examples) for details.

rjschwei:
Yes, that's the technical detail, question is, are you in favor of opening the range in this way?

bear454:
Well, yes.

If we're going to do exact version lockdowns, we need to add some automation to read those versions from OBS.

I think that's overkill, when all we really want is to ensure that our stuff works with current stuff. Using the `~=` matcher lets us ignore minor changes to the dependencies (which *shouldn't* break anything, and we don't have time to babysit anyway), but forces us to make a change, and do some testing, on a major change to a dependency.

I think that's the best balance we can achieve with the tools at hand.

schaefi:
I'll took this one because I'm currently working on azurectl due to the $MS review of my SDK patches
James hope you don't mind

schaefi:
Fixed

PR #75, opened 2015-11-04
Update module dependencies strictness

schaefi:
Instead of requiring an exact version we opened this to
match the major version and its releases. This fixes Issue #73


schaefi:
Please note the tests will fail until #74 is merged

PR #78, opened 2015-11-23
Added azurectl compute image update

schaefi:
Allow to update OS image metadata

### DIFF
--- a/azurectl/compute_image_task.py
+++ b/azurectl/compute_image_task.py
@@ -20,6 +20,16 @@ usage: azurectl compute image -h | --help
            [--delete-disk]
        azurectl compute image replicate --name=<imagename> --regions=<regionlist> --offer=<offer> --sku=<sku> --image-version=<version>
        azurectl compute image unreplicate --name=<imagename>
+       azurectl compute image update --name=<imagename>
+           [--description=<description>]
+           [--eula=<eula>]
+           [--image-family=<image_family>]
+           [--icon-uri=<icon_uri>]
+           [--label=<label>]
+           [--language=<language>]
+           [--privacy-uri=<privacy_uri>]
+           [--published-date=<date>]
+           [--small-icon-uri=<small_icon_uri>]
        azurectl compute image publish --name=<imagename>
            [--private]
            [--msdn]
@@ -30,6 +40,16 @@ commands:
         list available os images for configured account
     create
         create OS image from VHD disk stored on blob storage container
+    delete
+        delete OS image from VM image repository
+    replicate
+        replicate registered VM image to specified regions
+    unreplicate
+        unreplicate registered VM image from specified regions
+    update
+        update OS image meta data
+    publish
+        publish registered VM image
     help
         show manual page for image command
 
@@ -58,6 +78,7 @@ options:
         restrict publish scope to the Microsoft Developer Network
 """
 # project
+from logger import log
 from cli_task import CliTask
 from azure_account import AzureAccount
 from data_collector import DataCollector
@@ -97,6 +118,8 @@ class ComputeImageTask(CliTask):
             self.__unreplicate()
         elif self.command_args['publish']:
             self.__publish()
+        elif self.command_args['update']:
+            self.__update()
 
     def __help(self):
         if self.command_args['help']:
@@ -168,3 +191,34 @@ class ComputeImageTask(CliTask):
             )
         )
         self.out.display()
+
+    def __update(self):
+        log.info(
+            'Updating image metadata for: %s', self.command_args['--name']
+        )
+        update_elements = {
+            'description':
+                self.command_args['--description'],
+            'eula':
+                self.command_args['--eula'],
+            'image_family':
+                self.command_args['--image-family'],
+            'icon_uri':
+                self.command_args['--icon-uri'],
+            'label':
+                self.command_args['--label'],
+            'language':
+                self.command_args['--language'],
+            'privacy_uri':
+                self.command_args['--privacy-uri'],
+            'published_date':
+                self.command_args['--published-date'],
+            'small_icon_uri':
+                self.command_args['--small-icon-uri']
+        }
+        for name, value in update_elements.iteritems():
+            if value is not None:
+                log.info('--> %s = %s', name, value)
+        self.image.update(
+            self.command_args['--name'], update_elements
+        )
--- a/.travis.script
+++ b/.travis.script
@@ -5,7 +5,7 @@ git clone https://github.com/SUSE/azure-sdk-for-python.git
 git clone https://github.com/SUSE/azure-storage-python.git
 
 # Install specified branch
-branch_name=compute_management_service
+branch_name=os_image_metadata
 (
     cd azure-sdk-for-python
     git checkout $branch_name
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -117,6 +117,10 @@ class AzureOsImagePublishError(AzureError):
     pass
 
 
+class AzureOsImageUpdateError(AzureError):
+    pass
+
+
 class AzureConfigParseError(AzureError):
     pass
 
--- a/azurectl/defaults.py
+++ b/azurectl/defaults.py
@@ -49,3 +49,11 @@ class Defaults(object):
                 'Specified region %s not in lookup table'
             )
         return azure_domain[region]
+
+    @classmethod
+    def get_attribute(self, instance, name):
+        return getattr(instance, name)
+
+    @classmethod
+    def set_attribute(self, instance, name, value):
+        return setattr(instance, name, value)
--- a/azurectl/image.py
+++ b/azurectl/image.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import collections
 from tempfile import NamedTemporaryFile
 from azure.servicemanagement import ServiceManagementService
 from azure.storage.blob import BlobService
@@ -23,8 +24,10 @@ from azurectl_exceptions import (
     AzureOsImageDeleteError,
     AzureOsImageReplicateError,
     AzureOsImageUnReplicateError,
-    AzureOsImagePublishError
+    AzureOsImagePublishError,
+    AzureOsImageUpdateError
 )
+from defaults import Defaults
 
 
 class Image(object):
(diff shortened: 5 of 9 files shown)


schaefi:
Also see my SDK patch here: https://github.com/SUSE/azure-sdk-for-python/tree/os_image_metadata

I have not yet created a PR for this

azurectl/compute_image_task.py
```diff
@@ -20,6 +20,16 @@
...
        azurectl compute image replicate --name=<imagename> --regions=<regionlist> --offer=<offer> --sku=<sku> --image-version=<version>
        azurectl compute image unreplicate --name=<imagename>
+       azurectl compute image update --name=<imagename>
+           [--description=<description>]
+           [--eula=<eula>]
+           [--image-family=<image_family>]
+           [--icon-uri=<icon_uri>]
+           [--label=<label>]
+           [--language=<language>]
+           [--privacy-uri=<privacy_uri>]
```
rjschwei: Some kind of explanation what privacy-uri is supposed to be would be great, either via --help or man page entry

azurectl/compute_image_task.py
```diff
@@ -166,3 +189,39 @@ def __publish(self):
             )
         )
         self.out.display()
+
+    def __update(self):
+        update_record = {}
```
rjschwei: We do a lot of work before checking if the image we want to modify actually exists, which happens in image.update(). It would be nice if the image object would have a method that can be called from here, image.exists(name) or maybe a get method, such that we could have:

if not image.get(name):
  raise ....

schaefi: don't understand that part. The implementation of the update() method in the Image class does that:

```python
try:
            os_image = service.get_os_image(name)
        except Exception as e:
            raise AzureOsImageUpdateError(
            )
```

schaefi: or do you mean an explicit check for the existence of the image ? While I could understand we needed this check for running an instance I think we can trust the exception from the SDK and don't need to add our own one in this case

azurectl/compute_image_task.py
schaefi: agreed and I have no idea what the privacy-uri is :) same applies to other parts of the image metadata e.g icon-uri. Thus it would be nice if we could add a reference link to a Microsoft page to let people point to that information. I did not found such an information on a quick search

azurectl/compute_image_task.py
rjschwei: On 11/29/2015 04:06 PM, Marcus Schäfer wrote:
>> @@ -166,3 +189,39 @@ def __publish(self):
>>              )
>>          )
>>          self.out.display()
>> +
>> +    def __update(self):
>> +        update_record = {}
> 
> or do you mean an explicit check for the existence of the image ? While I could understand we needed this check for running an instance I think we can trust the exception from the SDK and don't need to add our own one in this case

Yes an explicit check. My point is that "a lot of work" is done, i.e.
checking all the arguments and setting up a dictionary for an image that
may not exist. We should exit early, i.e. if the image does not exist,
exit without doing any work.



-- 
rjschwei@suse.com
IRC: robjo

azurectl/compute_image_task.py
schaefi: I could only found a reference to an online book

https://books.google.de/books?id=9-zsBAAAQBAJ&pg=PA66&dq=Azure+image+properties&hl=en&sa=X&ved=0ahUKEwjTssXor7jJAhVCBywKHfVtBP4Q6wEILzAA#v=onepage&q=Azure%20image%20properties&f=false

guess we don't want that in our docs ?

azurectl/compute_image_task.py
schaefi: This is what you get today

$ ./azurectl compute image update --name foo

```
ERROR: AzureOsImageUpdateError: 'AzureMissingResourceHttpError: Not Found\n<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>ResourceNotFound</Code><Message>The image with the specified name does not exist.</Message></Error>'
```

schaefi: So we could add a "better" error message by catching "AzureMissingResourceHttpError" but actually this is something I really like to avoid. The reason is that the error handling in the SDK uses the same raise type for different types of errors. Thus if we add our own intelligence here we cover the message from the SDK which could be a different error message

rjschwei: On 11/30/2015 09:43 AM, Marcus Schäfer wrote:
>> @@ -166,3 +189,39 @@ def __publish(self):
>>              )
>>          )
>>          self.out.display()
>> +
>> +    def __update(self):
>> +        update_record = {}
> 
> So we could add a "better" error message by catching "AzureMissingResourceHttpError" but actually this is something I really like to avoid. The reason is that the error handling in the SDK uses the same raise type for different types of errors. Thus if we add our own intelligence here we cover the message from the SDK which could be a different error message

No, it is not about the error message. It is about creating an early
exit condition.

Currently the code structure is such that we do work:

- setup an empty dictionary
- process arguments
- call another method on another object
- if the image doe not exist, complain

What I think would be nicer is
- if the image does not exist complain
- else setup an empty dictionary
- process arguments
- call another method on another object





-- 
rjschwei@suse.com
IRC: robjo

schaefi: sorry where do you see the creation of an empty directory ?

schaefi: the command processing can't be skipped, how would you know the image before processing the command arguments ??

schaefi: The current flow is:

1. command processing
2. create a service management instance
    this would be the first time we are able to check for the existence of the image in azure
3. run the update SDK call
    you get the error now if there is any

still don't see any benefit from changing that

rjschwei: On 11/30/2015 10:11 AM, Marcus Schäfer wrote:
>> @@ -166,3 +189,39 @@ def __publish(self):
>>              )
>>          )
>>          self.out.display()
>> +
>> +    def __update(self):
>> +        update_record = {}
> 
> sorry where do you see the creation of an empty directory ?

Right there, update_record = {} ???




-- 
rjschwei@suse.com
IRC: robjo

rjschwei: On 11/30/2015 10:13 AM, Marcus Schäfer wrote:
>> @@ -166,3 +189,39 @@ def __publish(self):
>>              )
>>          )
>>          self.out.display()
>> +
>> +    def __update(self):
>> +        update_record = {}
> 
> The current flow is:
> 
> 1. command processing
> 2. create a service management instance
>     this would be the first time we are able to check for the existence of the image in azure
> 3. run the update SDK call
>     you get the error now if there is any
> 
> still don't see any benefit from changing that

We are clearly missing each other. It is not worth father discussion,
just merge it.

-- 
rjschwei@suse.com
IRC: robjo

schaefi: sorry I don't wanted to displease you, I just did not understand your point and I'm still confused

I think for merging this it's too early. We are still missing a doc hint for the parts we allow to change. You already pointed that out and it is still valid

Also this is based on a change in the azure SDK where I did not got a feedback from $MS as well as from my team. If we want to merge this we have to make the decision to again carry a custom SDK patch. Do we want this yes/no

sorry if this sounds pedantic, but imho these are all valid points

rjschwei: On 11/30/2015 10:44 AM, Marcus Schäfer wrote:
>> @@ -166,3 +189,39 @@ def __publish(self):
>>              )
>>          )
>>          self.out.display()
>> +
>> +    def __update(self):
>> +        update_record = {}
> 
> sorry I don't wanted to displease you, I just did not understand your point and I'm still confused

Yes, we are still missing each other, but given that I am apparently
unable to explain my suggestion in a way that makes sense and that the
suggestion is not critically important I see no reason to continue the path.

> 
> I think for merging this it's too early. We are still missing a doc hint for the parts we allow to change.

Yes, but that can be as easy as adding something to the man page in a
follow up request, or integrated into this request.

> You already pointed that out and it is still valid
> 
> Also this is based on a change in the azure SDK where I did not got a feedback from $MS as well as from my team. If we want to merge this we have to make the decision to again carry a custom SDK patch. Do we want this yes/no

I think we should carry the SDK patch in OBS while we work with the
upstream maintainers of the SDK to get something that meets our needs
integrated.

> 
> sorry if this sounds pedantic, but imho these are all valid points

Very valid, I agree.

-- 
rjschwei@suse.com
IRC: robjo

azurectl/compute_image_task.py
bear454: * `description` is a long-format text description of the image, including limited HTML tag support.
* `eula` URL where an end-user may read the text of the image's end user license agreement.
* `image-family` an arbitrary classification of the image, in order to categorize versions of an image in the Azure gallery; this is replaced by a combination of `offer` and `sku` in the Azure marketplace.
* `icon-uri` either (a) the name of a 100x100px image file local to Azure gallery, or (b) the URL of a 100x100px image to use as a large icon in the Azure gallery list.
* `label` a one-line 'title' for the image
* `language` I'm not sure - I don't set it.
* `privacy-uri` URL where an end user may read the privacy policy that governs information sent to the authors when an image is used.

bear454: screenshot of some attributes

schaefi: great picture and explanation :) would be cool if a man page could show that.
I'm going to add the text information to the manual pages

schaefi:
I have added manual page data for the update command

schaefi:
JFI: I got a positive review of my SDK patch, however the unit test system is very complex. I will need some time to figure out the recording stuff

azurectl/compute_image_task.py
bear454: @rjschwei mentioned an empty *dictionary*. @schaefi asked about an empty *directory*.

schaefi:
guess I finally understood what Robert meant

schaefi:
I have tested changing several data but got confused. It is no problem to change e.g language or label or description to whatever we want. But try to change the image-family for example

$ azurectl compute image update --name SLES12-JeOS-for-MS-HyperV --image-family suse

no error is reported from Azure but the value is also not applied. Same for --eula

That's weird

bear454:
@schaefi In practice, we've used the same string for ImageFamily and Label, for example "SUSE Linux Enterprise Server 12 (Premium Image)" .

My assumption is that altering this attribute is only available via accounts that have been flagged as 'publishers', in our case, the 'updates' account.

schaefi:
ok that makes sense, however we will not receive any error message from the service. I'm updating the documentation and will also add a check if the element was really updated

schaefi:
Time to get something done here, I have created a patch for our SDK patches because I think it will still take some time until $MS will merge my patch due to the test recording problematic. However $MS will create the recording for me and they are fine with the patch. Thus it's just a matter of time and I'd like to  move one. Therefore I'll carry the SDK patch until it can be merged

PR #79, opened 2015-11-24
Added support for replicating to all regions

### LINKED ISSUE #77: allow to replicate to all regions
The following is desired

```
azureclt compute replicate --name=foo --regions=all
```

schaefi:
The option --regions supports the value 'all', this Fixes #77


rjschwei:
LGTM

Issue #80, opened 2015-11-24
missing dependency on python-setuptools


bear454:
Upon installation on a clean system, first run outputs:

```
> azurectl 
Traceback (most recent call last):
  File "/usr/bin/azurectl", line 5, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named pkg_resources
```

`pkg_resources` is provided by python-setuptools.

rjschwei:
I am not fond of having setuptools as a runtime dependency. Makes me wonder if we should not change our approach and instead of letting setuptools  do the generation that triggers this dependency write the main entry point such that setuptools is not needed at runtime?

bear454:
@rjschwei , @schaefi : now we have two different solutions, both with and without setuptools. Comments, please?

schaefi:
as said in the pull request, I'd like to see us using python-setuptools

schaefi:
Fixed

PR #81, opened 2015-11-24
Add missing dependency

bear454:
For issue #80


schaefi:
Thanks, could you also add 'setuptools~=5.4' to install_requires in setup.py

PR #82, opened 2015-12-01
No runtime dependency on setuptools

bear454:
Use the bin/azurectl script instead of generating a cmd line entry point via
setuptools.

re: https://github.com/SUSE/azurectl/issues/80#issuecomment-160423404


bear454:
This is an alternative solution to PR #81, to resolve Issue #80.

rjschwei:
Principally this works for me, although I think the redirection should not be necessary. This implementation runs a bash process which then runs a python process. The script can be implemented in python and then we woudl only have to spawn one process to get azurectl working.

bear454:
Understood, and I agree; I just wanted the shortest route to a possible solution, until there's consensus on the solution. At that point, I'm fine with refining either PR; just not both ;-)

schaefi:
Actually I don't understand why we don't want to use python-setuptools. This is the "pythonic" way of adding an entry point into the application and I really like this concept. I would prefer it from writing an extra script just for removing a dependency.

I have checked if we are missing python-setuptools in any of the target distributions and it exists everywhere:

```
osc search:

SUSE:SLE-11:GA                                        python-setuptools
SUSE:SLE-12-SP1:GA                                python-setuptools
SUSE:SLE-12:GA                                        python-setuptools
....
```

So what's wrong with the first pull request from James ? I would just go with that one

rjschwei:
It is not a question about setuptools being available. Also you are correct that a number of projects generate entry point with setuptools just like this.

My opinion is that there is a design flaw w.r.t. the dependencies. A tool designed and implemented for build and installation should not really generate a run time dependency onto itself. We can draw 2 parallels in the C world for example. Everyone would complain if the end user would have to install make to run anything that uses make as its build engine. Also everyone would complain if people would have to install the compiler (gcc) just because something is written in C/C++.

I agree that in interpreted languages, the separation of concerns is not as strong as for compiled languages. That however does not necessarily imply that we have to follow suit.

That said I was just raising the question. Personally I am not in favor of having a runtime dependency on setuptools. However if others think we should continue with that approach I certainly am not going to run screaming from the room. BTW it is not necessary to have an extra script, as can be seen with the ec2utils implementation it is possible to do all of this without setuptools generated entry points without additional moving parts.

It is a personal preference thing. I lean toward no runtime dependency on setuptools, but so far I have contributed very little code to this specific project thus if others prefer to keep the runtime dependency then so be it.

schaefi:
> BTW it is not necessary to have an extra script, as can be seen with the ec2utils implementation it is
> possible to do all of this without setuptools generated entry points without additional moving parts.

The programs in the Enceladus/ec2utils directory are written in a different way. You are providing the application as an executable e.g 'ec2deprecateimg' without the .py extension and install this to the desired place. That way you don't need an extra entry point into the application this is correct.

If this should be changed you need to change the project structure of azurectl. If not one script to run the application is required, needs to be written and maintained.  Using setuptools avoids this but creates a dependency on the system, true

if I got it correctly your concern with this is that people will complain when they have to install python-setuptools in order to run azurectl. For me personally this is not a problem, even as a user I would not have had a problem to have the setuptools on my system. They are light weight without any further dependencies since the python core stack already provides most of the components.

With compiled languages the requirement would pull in an entire development stack, headers, libraries, tools, and I agree that's what people don't like, but for python it's only setuptools, roughly 1MB in size and most probably already on the system because _some_ of the many python packages will most probably already need it

I'm sorry I wouldn't be bothered by setuptools and they made my life a little bit easier

rjschwei:
Fair enough just add the setuptools runtime dependency

schaefi:
thanks, going to close this one then, merge in James other PR, update setup.py and the version and will submit a new package for testing

PR #84, opened 2016-01-16
Add command for listing a single image, by name.

bear454:
... because I'm tired of grep'ing the results.

### DIFF
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -205,6 +205,10 @@ class AzureOsImageListError(AzureError):
     pass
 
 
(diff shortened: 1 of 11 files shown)


bear454:
@schaefi - image tests passed, and all tests pass locally. Are the travis failures due to our patched SDK version?

schaefi:
patch looks great to me :+1: 

With regards to the failed tests. This is because travis seems to have changed the pep8 version or policy to be more strict now with compare statements. In the tests I often used "== False" and not
"is False". They now complain about this.

I have fixed the tests and you should be able to simply rebase your work which should make everything green

PR #88, opened 2016-01-21
Fixed check for URI based OS image attributes

### LINKED ISSUE #86: on `compute image update`, privacy uri may change, suggesting failure
```
$ azurectl --account publish compute image --name test-attributes update --privacy-uri http://privacy.uri
INFO: Updating image metadata for: test-attributes
INFO: --> privacy_uri = http://privacy.uri
ERROR: AzureOsImageUpdateError: 'The element(s) "privacy_uri" could not be updated. Please check if your account is registered as image publisher'
```
Yet, viewing the data shows that the attribute *is* assigned, just modified:
```
$ azurectl --account publish compute image --name test-attributes show
{
  "image": {
    ...
    "privacy_uri": "http://privacy.uri/",
    ...
  }
}
```

In the above example, the API appended an ending '/', causing it to fail `azurectl`'s validation.

schaefi:
After an update of e.g privacy_uri the API might have changed the value to be still correct but different from the original value provided on the commandline by e.g a slash (/). Fixes #86

### DIFF
--- a/azurectl/image.py
+++ b/azurectl/image.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import os
 import collections
 from tempfile import NamedTemporaryFile
 from azure.servicemanagement import ServiceManagementService
@@ -192,6 +193,10 @@ class Image(object):
         for name in os_image_attributes:
             value_desired = Defaults.get_attribute(os_image, name)
             value_current = Defaults.get_attribute(os_image_updated, name)
+            if '_uri' in name:
+                # The API changes the provided path value. Thus we normalize
+                value_desired = os.path.normpath(value_desired)
+                value_current = os.path.normpath(value_current)
             if value_desired != value_current:
                 elements_not_changed.append(name)
         if elements_not_changed:
--- a/azurectl/image.py
+++ b/azurectl/image.py
@@ -12,7 +12,7 @@
 # limitations under the License.
 #
 import os
-import time
+import dateutil.parser
 import collections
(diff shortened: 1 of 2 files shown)


schaefi:
Sorry for addressing two issues in the same pull request but this is because I could not separate them easily without creating conflicts

schaefi:
you are right, we should be more user friendly :smile: how about it now ?

bear454:
Awesome, Schaefi!

azurectl/image.py
```diff
@@ -255,3 +220,53 @@ def publish(self, name, permission):
...
+            'os': image.os,
+            'os_state': image.os_state,
+            'pricing_detail_link': image.pricing_detail_link,
+            'privacy_uri': image.privacy_uri,
+            'published_date': image.published_date,
+            'publisher_name': image.publisher_name,
+            'recommended_vm_size': image.recommended_vm_size,
+            'show_in_gui': image.show_in_gui,
+            'small_icon_uri': image.small_icon_uri
+        }
```
rjschwei: While this is nice and explicit it is also very long and I am not certain that there is value in listing all these attributes one line at a time. How about this?

decorator = []
for image_attr in dir(image):
    if '__' in image_attr:
        continue
    value = getattr(image, image_attr)
    if inspect.ismethod(value):
       continue
   decorator[image_attr] = value

The advantage is that we do not have to list all these attributes explicitly and when a new attribute shows up it goes into the decorator automatically. The disadvantage is that is a new attribute on image is needed that is not supposed to be in the decorator it has to start with "__" of course that test could easily be modified to also include some oher prefix.

PR #90, opened 2016-01-22
- Shorten the implementation of __decorate_image_for_results

rjschwei:
+ There is no value in creating an explicit list for a dictionary
    initialization. Also using explicit naming opens the door to us
    forgetting a new value that may be added later.


schaefi:
Hmm, yes this patch makes the code smaller and is smart, but we loose a lot of information which is now hidden.
I'm not a friend of changing from explicit to implicit code. This also goes in line with the python zen
see here: https://www.python.org/dev/peps/pep-0020/

I think explicitly adding new values when they come up makes changes more clear. The magically addition of attributes just because they appear in an object instance open a pool of potential problems and is hard to debug and also makes testing harder.

I was more in favor with James private decorator implementation

Thoughts ?

rjschwei:
Well I had added this as a discussion point to another pull request that at the same time was closed by James. It was just a suggestion and intended for discussion there.

I agree that maintaining things explicitly in general is a good thing. IN this case the explicit instantiation of the dictionary is rather lengthy an unwieldy IMHO, but is certainly valid.

Issue #91, opened 2016-02-02
Uploads die upon failure; should recover


bear454:
```
$ azurectl --account publish compute storage upload --name SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed --source binaries/SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed.xz
Uploading: [                                        ] 0%ERROR: AzureStorageUploadError: 'error: [Errno 104] Connection reset by peer'

$ azurectl --account publish compute storage upload --name SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed --source binaries/SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed.xz
Uploading: [#############                           ] 33%ERROR: AzureStorageUploadError: "SSLError: ('The write operation timed out',)"

$ azurectl --account publish compute storage upload --name SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed --source binaries/SUSE-Manager-Server.x86_64-2.1.11-Build1.46.vhdfixed.xz
Uploading: [#####################################   ] 93%ERROR: AzureStorageUploadError: "SSLError: ('The write operation timed out',)"
```
In each of these cases, the 'slice' of data being uploaded could have been retried, instead of failing out the whole upload.

schaefi:
>    In each of these cases, the 'slice' of data being uploaded could have
>    been retried, instead of failing out the whole upload.

right, that would be a nice thing to add. We don't have a retry feature

schaefi:
done

PR #92, opened 2016-02-03
Don't die so easily when uploading

bear454:
Uploading a 30GB image, in 4MB chunks, can be fragile. Instead of giving up and raising an exception the first time a block fails, retry a few times.

re: Issue #91

### DIFF
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -221,6 +221,18 @@ class AzurePageBlobAlignmentViolation(AzureError):
     pass
 
 
+class AzurePageBlobSetupError(AzureError):
+    pass
+
+
+class AzurePageBlobZeroPageError(AzureError):
+    pass
+
+
+class AzurePageBlobUpdateError(AzureError):
+    pass
+
+
 class AzureRequestTimeout(AzureError):
     pass
 
@@ -242,6 +254,10 @@ class AzureServiceManagementError(AzureError):
 
 
 class AzureStorageListError(AzureError):
+    pass
+
+
+class AzureStorageStreamError(AzureError):
     pass
 
 
--- a/azurectl/page_blob.py
+++ b/azurectl/page_blob.py
@@ -1,0 +1,146 @@
+# Copyright (c) 2015 SUSE Linux GmbH.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


bear454:
@schaefi while all the tests pass with the code change, I feel this needs a more unique test. I'd appreciate your help constructing something we could use to unit test the change.

As far as real-world testing goes, I pulled my network cable out during a test, waited for a timeout, and put it back in. Upload finished!

schaefi:
Thanks for the patch, this looks good and surely works. However I think the code is now relatively complex and I suggest to refactor it which should also make it more testable. 

What I have in mind could be expressed best in code I guess. Based on your branch I would add a new iterator class for page blobs. That will make the code easier to read and we are able to mock
junks of stream data in order to really test a single iteration and its expected behaviour. Also the
retry code fits best into the iterator

https://w3.suse.de/~ms/0003-Refactor-page-blob-upload.patch

This is just an idea

Let me know what you think

bear454:
@schaefi agreed; I woke up this morning and thought "it's spaghetti; let's refactor it."

I didn't have in mind to build an iterator class, just a few private methods, but you're approach is probably more maintainable in the long run.  Add it to the PR, and let's see what else needs to be done here.

schaefi:
ok will do

schaefi:
all right that should do the trick

bear454:
@schaefi:
```
Uploading: [####################################### ] 97%ERROR: Unexpected error:
  File "<string>", line 1, in <module>
  File "azurectl/azurectl.py", line 29, in main
    App()
  File "azurectl/app.py", line 28, in __init__
    task_class().process()
  File "azurectl/compute_storage_task.py", line 157, in process
    self.__upload()
  File "azurectl/compute_storage_task.py", line 195, in __upload
    self.__upload_with_progress()
  File "azurectl/compute_storage_task.py", line 211, in __upload_with_progress
    self.__process_upload()
  File "azurectl/compute_storage_task.py", line 224, in __process_upload
    self.command_args['--max-chunk-size']
  File "azurectl/storage.py", line 70, in upload
    stream.close()
AttributeError: 'XZ' object has no attribute 'close'
```

schaefi:
yes because I'm stupid and removed the close and the test from the XZ class.... because I thought I wouldn't need to explicitly call it. Shame_on_me

fixing

schaefi:
Fixed now, commit e1de7cec988fd40928cc8a8de688f3239fd65d06 does what I had intended

bear454:
:D

PR #94, opened 2016-02-19
More disconnected regions

bear454:
A few things that needed to be added for complete support of disconnected regions*.

* Mooncake and Black Forest; I don't have access to Fairfax.


schaefi:
yay nice

PR #95, opened 2016-02-22
Commands for managing public IP address reservations

bear454:


azurectl/app.py
```diff
@@ -23,6 +24,6 @@ def __init__(self):
         app = CliTask(should_load_config=False)
         action = app.cli.get_command()
         service = app.cli.get_servicename()
-        task_class_name = service.title() + action.title() + 'Task'
+        task_class_name = camelize(service) + camelize(action) + 'Task'
```
schaefi: that's nice :+1:

schaefi:
I'm impressed, that's great work :smile:

merge it :+1:

azurectl/app.py
bear454: Except I need to package that dependency :/

schaefi: >    Except I need to package that dependency :/

Oh this is not core python ? In this case I think we should stick with
the old code

bear454: I'll write a local version of camelize instead. We really need something more robust than title... as using just title, we end up with ComputeReserved_IpTask .

bear454:
If you don't mind; I'm going to refactor 'reserved_ip' to 'reserved-ip' in the commands, to be more consistent.

schaefi:
looks all good to me, thanks for not introducing a new package dependency

PR #96, opened 2016-03-03
Added compute storage container create|delete

schaefi:
Allows creation and deletion of containers for the
configured storage account


schaefi:
Uff, azure-storage 0.30 as released on pypi is again incompatible with 0.20.3

I really dislike that $MS habit to not care for compatibility, guess that's a hard wired in the culture of a closed source company.

I'll adapt our code base and put the refactoring commit to this branch

schaefi:
Oh well the last commit changes many things. This needs intensive testing first, please do not merge yet

schaefi:
I have tested storage and servicemanagement operations, so far they are all functional

The update of azure storage from 0.20 to 0.30 offers many new blob operations. The update is required in order to support creation/deletion of containers which was the original intention of this
pull request.

However the update also fundamentally changes the way how data is uploaded. The sdk is now based on blocklists and methods like put_blob or put_page as we have used them are gone from the public interface. However they introduced 'create_blob_from_stream' which basically does all the stuff we had implemented ourselves with chunking and progress information. Therefore I make use of this method now and deleted our code.

I have done some test uploads which worked for me. I did not investigate if the optimization of zero pages as we had it is used by them too

schaefi:
@bear454 I'm going to merge the changes to unblock you with the storage account patches.

I have tested uploading and other azurectl calls and did not stumble over regressions
last famous words :smile:

rjschwei:
Thanks, for following the bouncing ball of incompatible changes. All of this is certainly turning out to be a much bigger pain than we originally had in mind.

Going forward it would be great to see code changes that are necessary for new upstream SDK versions and new features implemented in separate pull requests. That will make review of the new features easier for others.

schaefi:
>    Thanks, for following the bouncing ball of incompatible changes. All of
>    this is certainly turning out to be a much bigger pain than we
>    originally had in mind.

right

>    Going forward it would be great to see code changes that are necessary
>    for new upstream SDK versions and new features implemented in separate
>    pull requests. That will make review of the new features easier for
>    others.

I agree and I would have done it if I just would have known how many
incompatible changes the storage api came up with. Unfortunately it
has turned into a big change on our side.

This is a learning process when working with $MS, I don't expect
any new version to be compatible with the former. Thus I suggest
to add a 

* check for SDK compatibility

task to any of our azurectl cards _before_ we start adding new
features.

The problem with azurectl and the dependency to the $MS SDK will
become even bigger in the future. As you might know our entire
service management is based on what they call "legacy service management"
Their new initiative offers a complete new way of authentication
and the SDK for service management is a new code stream for
resource(token) based auth vs. key based auth. If you look into the
nodejs code you can already see that they doubled the implementation
depending on the authentication method.

At the time they switch off support for key based auth. we are
screwed and can start from scratch with azurectl and all its service
management based tasks

In summary I agree to what you have said, but it's not so easy, at
least I realize such problems almost any time when I just wanted to
add some new stuff with the expectation that an SDK should have a
stable ABI...

PR #97, opened 2016-03-08
Storage account management commands

bear454:


azurectl/storage_account.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class MinimalContainerAccount(object):
+    """
+        The Minimum required to instantiate Container
+    """
+    def __init__(self, storage_name, storage_key, blob_service_host_base):
+        self.storage_name = lambda: storage_name
+        self.storage_key = lambda: storage_key
+        self.get_blob_service_host_base = lambda: blob_service_host_base
```
schaefi: why lambda expressions ?

azurectl/asm_object.py
```diff
@@ -0,0 +1,44 @@
...
+import collections
+from tempfile import NamedTemporaryFile
+from azure.servicemanagement import ServiceManagementService
+# project
+from azurectl_exceptions import *
+from defaults import Defaults
+
+
+class AsmObject(object):
+
```
schaefi: AsmObject probably means Azure Service Management Object, or Account Service Management Object
correct ?

I think the name of the class is not explicit enough, given that all Azure classes uses ServiceManagement I would not search for 'sm' or 'asm'. I also think the class name should not contain Object as part of its name. I was confused by an object thing inheriting from object :smile: 

If we can't find a better description for the responsibility of the class we should at least have doc string describing it
maybe a better name would be:

* AccountService
* AcountServiceManagement

just an idea

azurectl/storage_account.py
bear454: `Container` expects an `Account` that responds to specific functions. Lambdas seemed like the cleanest way to implement `MinimalContainerAccount#storage_name()`, etc., which return the expected values.

azurectl/asm_object.py
bear454: Renamed `AsmObject` to `ServiceManager` in 4b079ed .

bear454: I want to move more to a model like this; we have a lot of repetitive boilerplate that can be handled more cleanly through inheritance.

azurectl/storage_account.py
schaefi: I use lambdas to produce closures or abstract code blocks into a callable namespace
I do not understand the value behind:

* container_account.storage_name

versus

* container_account.storage_name()

If you are trying to create access methods for instance variables I would be more in favour to have them as function definition instead of a lambda. I know in ruby we have "attr_reader" and everything should be a method but in python you would need to create a function or lambda for it which seems not pythonic. I'm not a python expert but creating lambdas for static values feels weird

Will ask the python Gurus for the pros and cons about this construction

bear454: I have to instantiate Container with an Account - the Account used may not be the appropriate context, as I'm looking up information about storage accounts that may not be the storage account selected in config.

So, I can't simply pass `self.account` on ... I need to generate something that satisfies Container's expectations for account - which include calling `self.account.storage_name()`, `self.account.storage_key()`, and `self.account.get_blob_service_host_base()`. Simply having the attributes present fails, as you can't execute a string:

```
>>> foo = "bar"
>>> foo
'bar'
>>> foo()
  File "<stdin>", line 1, in <module>
TypeError: 'str' object is not callable
```
So there needs to be some 'callable' context. I didn't want to rewrite Account, as we use it's behavior in every class, and I didn't want to change Container, as it works fine the way it is, so I made up the smallest possible shim to get the behavior I wanted - hence the class name.

schaefi: Ah ok now I understand why you need this. I hope you don't mind but I find it confusing to have a MinimalContainerAccount class in the code dealing with storage_account. The use of lambda here is also only required because you need a callable because of the Container constructor. What do you think of the following solution:

```
diff --git a/azurectl/container.py b/azurectl/container.py
index 995943b..51387ad 100644
--- a/azurectl/container.py
+++ b/azurectl/container.py
@@ -113,3 +113,14 @@ class Container(object):
         return 'https://{}.blob.core.windows.net/{}?{}'.format(
             self.account_name, container, signed_query
         )
+
+
+class ContainerService(Container):
+    """
+        Information from Azure storage containers.
+        Instance constructed from required storage account data
+    """
+    def __init__(self, storage_name, storage_key, blob_service_host_base):
+        self.account_name = storage_name
+        self.account_key = storage_key
+        self.blob_service_host_base = blob_service_host_base
diff --git a/azurectl/storage_account.py b/azurectl/storage_account.py
index 29aabcb..6391f52 100644
--- a/azurectl/storage_account.py
+++ b/azurectl/storage_account.py
@@ -14,6 +14,7 @@
 # project
 from service_manager import *
 from container import Container
+from container import ContainerService
 
 
 class StorageAccount(ServiceManager):
@@ -108,7 +109,7 @@ class StorageAccount(ServiceManager):
         return result
 
     def __add_containers_to(self, result):
-        container_account = MinimalContainerAccount(
+        container_account = ContainerService(
             result.service_name,
             result.storage_service_keys.primary,
             self.account.get_blob_service_host_base()
@@ -145,13 +146,3 @@ class StorageAccount(ServiceManager):
                 'secondary': result.storage_service_keys.secondary
             }
         return decorated
-
-
-class MinimalContainerAccount(object):
-    """
-        The Minimum required to instantiate Container
-    """
-    def __init__(self, storage_name, storage_key, blob_service_host_base):
-        self.storage_name = lambda: storage_name
-        self.storage_key = lambda: storage_key
-        self.get_blob_service_host_base = lambda: blob_service_host_base
```

I still think the class name to construct a container with different constructor parameters could be improved but it should show the idea. Unfortunately python does not have the concept of methods with the same name but different parameter lists

Thoughts ?

azurectl/storage_account.py
```diff
@@ -0,0 +1,157 @@
...
+        regenerate_primary_key=False,
+        regenerate_secondary_key=False
+    ):
+        try:
+            current = self.service.get_storage_account_properties(name)
+            self.service.update_storage_account(
+                name,
+                (description or current.storage_service_properties.description),
+                (label or current.storage_service_properties.label),
+                account_type=(account_type or current.storage_service_properties.account_type)
```
schaefi: The line is quite long, flake8 only likes this because we allow it for the docopt definitions. Would you mind to shorten this and the following ones to the 80 chars boundary ? Thanks

azurectl/storage_account.py
bear454: In that case, why not just make the init conditional?

```
diff --git a/azurectl/azurectl_exceptions.py b/azurectl/azurectl_exceptions.py
index 09d4f4f..9774180 100644
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -97,6 +97,10 @@ class AzureBlobServicePropertyError(AzureError):
     pass
 
 
+class AzureCannotInit(AzureError):
+    pass
+
+
 class AzureOsImageCreateError(AzureError):
     pass
 
diff --git a/azurectl/container.py b/azurectl/container.py
index 995943b..5ebec7a 100644
--- a/azurectl/container.py
+++ b/azurectl/container.py
@@ -29,10 +29,26 @@ class Container(object):
     """
         Information from Azure storage containers
     """
-    def __init__(self, account):
-        self.account_name = account.storage_name()
-        self.account_key = account.storage_key()
-        self.blob_service_host_base = account.get_blob_service_host_base()
+    def __init__(
+        self,
+        account=None,
+        account_name=None,
+        key=None,
+        blob_service_host_base=None
+    ):
+        if account:
+            self.account_name = account.storage_name()
+            self.account_key = account.storage_key()
+            self.blob_service_host_base = account.get_blob_service_host_base()
+        elif (account_name and key and blob_service_host_base):
+            self.account_name = account_name
+            self.account_key = key
+            self.blob_service_host_base = blob_service_host_base
+        else:
+            raise AzureCannotInit('''
+                Either an account, or account_name, key, and service host base
+                are required.
+            ''')
 
     def list(self):
         result = []
diff --git a/azurectl/storage_account.py b/azurectl/storage_account.py
index 29aabcb..ef00435 100644
--- a/azurectl/storage_account.py
+++ b/azurectl/storage_account.py
@@ -108,12 +108,12 @@ class StorageAccount(ServiceManager):
         return result
 
     def __add_containers_to(self, result):
-        container_account = MinimalContainerAccount(
-            result.service_name,
-            result.storage_service_keys.primary,
-            self.account.get_blob_service_host_base()
-        )
-        result.containers = Container(container_account).list()
+        container_args = {
+            'account_name': result.service_name,
+            'key': result.storage_service_keys.primary,
+            'blob_service_host_base': self.account.get_blob_service_host_base()
+        }
+        result.containers = Container(**container_args).list()
         return result
 
     def __decorate(self, result):
```

azurectl/storage_account.py
bear454: Fixed in 06cb44d.

azurectl/container.py
```diff
@@ -29,10 +30,26 @@ class Container(object):
...
+            self.blob_service_host_base = account.get_blob_service_host_base()
+        elif (account_name and key and blob_service_host_base):
+            self.account_name = account_name
+            self.account_key = key
+            self.blob_service_host_base = blob_service_host_base
+        else:
+            raise AzureCannotInit('''
+                Either an account, or account_name, key, and service host base
+                are required.
+            ''')
```
schaefi: all right let's go that route

PR #98, opened 2016-03-08
Add `compute shell` command -> drops to interactive python

bear454:
I find myself hacking this function in over and over, in order to implement new functionality. It makes more sense to just do it once as a command.

### DIFF
--- a/azurectl/compute_shell_task.py
+++ b/azurectl/compute_shell_task.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2016 SUSE.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


azurectl/compute_shell_task.py
```diff
@@ -0,0 +1,56 @@
...
+            account.publishsettings().subscription_id,
+            cert_file.name,
+            account.publishsettings().management_url
+        )
+
+        print """
+An instance of azure.servicemanagement.ServiceManagementService has been
+instantiated using the supplied credentials, as `service`.
+
+When you're finished, exit with `exit()`
```
schaefi: You can use dedent to make that info part nicer

```
from textwrap import dedent

session_message = dedent('''
    An instance of azure.servicemanagement.ServiceManagementService has been
    instantiated using the supplied credentials, as `service`.

    When you're finished, exit with `exit()` 
''').strip()

print session_message
```

schaefi:
Great idea :+1: 

just a small comment on the message text

azurectl/compute_shell_task.py
bear454: Changed in b1f9108.

PR #100, opened 2016-03-10
Create a uniq disk identifier on vm create

### LINKED ISSUE #99: Can't create two VMs from the same image in the same storage container
When instantiating a new VM, we create a disk file with the name '<image-name>_instance', in the selected storage container.

Since storage files have unique names, only one VM can created from the same image in a selected region and container.

See:

* https://github.com/SUSE/azurectl/blob/master/azurectl/compute_vm_task.py#L152
* https://github.com/SUSE/azurectl/blob/master/azurectl/virtual_machine.py#L137

schaefi:
When instantiating a new VM, we create a disk file with the
name '_instance', in the selected storage container. Since
storage files have unique names an additional uniq id information
is required to allow creation of multiple vm's from the same image
in a selected region and container. This fixes #99

### DIFF
--- a/azurectl/virtual_machine.py
+++ b/azurectl/virtual_machine.py
@@ -134,7 +134,13 @@ class VirtualMachine(object):
             )
 
         media_link = self.storage.make_blob_url(
-            self.container_name, disk_name + '_instance'
+            self.container_name, ''.join(
+                [
+                    cloud_service_name,
+                    '_instance_', system_config.host_name,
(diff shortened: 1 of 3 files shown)


schaefi:
ok I was concerned about a length limitation, I'm going to change it

schaefi:
I stumbled over a small problem in one of the unit test, this is just a one liner fix which I think is ok to handle here too

PR #104, opened 2016-03-12
Add bumpversion config

### LINKED ISSUE #101: Use python-bumpversion to handle versions
https://help.github.com/articles/creating-releases/

We want to follow this approach, bumpversion helps to do that

```
.bumpversion.cfg

[bumpversion]
current_version = 1.7.0
commit = True
tag = True

[bumpversion:file:azurectl/version.py]
```

```
$ bumpversion pathc|minor|major
$ git push --tags
```

will update the version and also create a git tag. That way we can follow the git release process

bear454:
This resolves #101 .

### DIFF
--- a/azurectl/version.py
+++ b/azurectl/version.py
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '1.8.0'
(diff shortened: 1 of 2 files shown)


schaefi:
yes that's nice, thanks

Issue #103, opened 2016-03-10
Cannot manage data disks


bear454:
Azurectl should have functions for creating empty data disks, mounting uploaded data disks, listing disks, and deleting disks.

https://trello.com/c/QyUM4wxB

schaefi:
done

PR #105, opened 2016-03-12
Commands for managing data disks (non-root/boot disks) on VMs.

bear454:
Create empty disks, show and delete individual disks, and list all disks on a VM.

Resolves issue #103

### DIFF
--- a/azurectl/data_disk.py
+++ b/azurectl/data_disk.py
@@ -12,10 +12,10 @@
 # limitations under the License.
 #
 from azure.common import AzureMissingResourceHttpError
+from azure.storage.blob.pageblobservice import PageBlobService
 from datetime import datetime
 # project
 from service_manager import *
-
 LUNS = 16  # there are 16 possible luns, numbered 0..15
 
 
@@ -155,14 +155,14 @@ class DataDisk(ServiceManager):
         )
 
     def __data_disk_url(self, filename):
-        return (
-            'https://' +
-            self.account_name +
-            '.blob.' +
-            self.account.get_blob_service_host_base() + '/' +
-            self.account.storage_container() + '/' +
-            filename
+        blob_service = PageBlobService(
+            self.account_name,
+            self.account_key,
+            endpoint_suffix=self.account.get_blob_service_host_base()
         )
+        return blob_service.make_blob_url(
+            self.account.storage_container(),
+            filename)
 
     def __decorate(self, data_virtual_hard_disk):
         return {
--- a/azurectl/data_disk.py
+++ b/azurectl/data_disk.py
@@ -83,6 +83,27 @@ class DataDisk(ServiceManager):
             )
         return self.__decorate(result)
 
+    def delete(
+        self,
+        cloud_service_name,
+        instance_name,
+        lun,
+        role_name=None
+    ):
+        try:
+            result = self.service.delete_data_disk(
+                cloud_service_name,
+                instance_name,
+                (role_name or cloud_service_name),
+                lun,
+                delete_vhd=True
+            )
+        except Exception as e:
+            raise AzureDataDiskDeleteError(
+                '%s: %s' % (type(e).__name__, format(e))
+            )
+        return result.request_id
+
     def get_first_available_lun(
         self,
         cloud_service_name,
--- a/azurectl/data_disk.py
+++ b/azurectl/data_disk.py
@@ -63,6 +63,26 @@ class DataDisk(ServiceManager):
             )
         return result.request_id
 
+    def show(
+        self,
+        cloud_service_name,
+        instance_name,
+        lun,
+        role_name=None
+    ):
+        try:
+            result = self.service.get_data_disk(
+                cloud_service_name,
+                instance_name,
+                (role_name or cloud_service_name),
+                lun
+            )
+        except Exception as e:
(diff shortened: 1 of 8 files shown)


azurectl/data_disk.py
```diff
@@ -0,0 +1,175 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from azure.common import AzureMissingResourceHttpError
+from datetime import datetime
+# project
+from service_manager import *
```
schaefi: to follow the explicit is better than implicit patter we should better write

```
from service_manager import ServiceManager
```

azurectl/data_disk.py
```diff
@@ -0,0 +1,175 @@
...
+                )
+            except AzureMissingResourceHttpError:
+                return lun
+            else:
+                lun += 1
+        raise AzureDataDiskNoAvailableLun(
+            "All LUNs on this VM are occupied."
+        )
+
+    def generate_filename(self, instance_name):
```
schaefi: Do you want this method to be in the public interface for this class ?

azurectl/data_disk.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __data_disk_url(self, filename):
+        return (
+            'https://' +
+            self.account_name +
+            '.blob.' +
+            self.account.get_blob_service_host_base() + '/' +
+            self.account.storage_container() + '/' +
+            filename
+        )
```
schaefi: iirc this generates a blob url. The storage sdk BaseBlobService provides a make_blob_url method which we could use

```python
make_blob_url(self.account.storage_container(), filename)
```

This imho should produce the same result (untested though) I have used the method in virtual_machine.py

schaefi:
Overall looks great, just some minor comments

Thanks

azurectl/data_disk.py
bear454: I'm trying to use this as a time/code saving device (and hopefully reduce my mistakes as well); `service_manager` defines a set of imports that can be reused; explicitly importing only `ServiceManager` means duplicating all the other imports, to no benefit.

azurectl/data_disk.py
bear454: Why not?

azurectl/data_disk.py
bear454: ah thanks; I'll use this instead

azurectl/data_disk.py
schaefi: I understand that it save a little bit of time but didn't we had the agreement to have no global imports except in tests ? I also see that service_manager imports all from azurectl_exceptions with '*' but does not raise any exception. We should either allow global imports or not but the mix is not a good idea

actually I wonder why flake8 did not complain about it

azurectl/data_disk.py
schaefi: I'm just asking where else you would see the need to create a data disk file name except for the data_disk class itself. It feels like this name is pretty exclusive to data_disk and doesn't need to be part of the public interface of the class

azurectl/data_disk.py
bear454: Replaced in 41cd85aa.

azurectl/data_disk.py
bear454: I've marked this and another helper private, per your suggestion, in 0a9befae.

In general though, I don't agree with making private methods, especially in code you wish to test. We're not dealing with a huge API footprint on any of these classes, and there's no intention to deliver this as a library; only a module supporting a CLI tool. So, in essence, the only real change I see here is adding a lot of boilerplate to the tests. :/

azurectl/data_disk.py
bear454: > actually I wonder why flake8 did not complain about it

I'm not running flake8; just the pep8 call from `tools/run-pep8`

bear454: > I understand that it save a little bit of time but didn't we had the agreement to have no global imports except in tests ? 

My goal is saving more that "a little bit of time"; we have huge amounts of repeated code; it's unnecessary and error-prone. For example:

```
    def __init__(self, account):
        self.account = account
        self.container_name = account.storage_container()
        self.account_name = account.storage_name()
        self.account_key = account.storage_key()
        self.cert_file = NamedTemporaryFile()
        self.publishsettings = self.account.publishsettings()
        self.cert_file.write(self.publishsettings.private_key)
        self.cert_file.write(self.publishsettings.certificate)
        self.cert_file.flush()

        self.service = ServiceManagementService(
            self.publishsettings.subscription_id,
            self.cert_file.name,
            self.publishsettings.management_url
        )
```
is repeated, with only slight differences, in `cloud_service.py`, `image.py`, `reserved_ip.py`, `storage.py`, and `virtual_machine.py`.

How may different places/ways to we instantiate `azure.servicemanagement.ServiceManagementService` ?

At some point it will be unmaintainable to have so much repetition, and Inheritance is a straightforward design pattern for eliminating repetition.

schaefi:
This is great work :smile: Thanks much

PR #107, opened 2016-03-14
Move to pytest

schaefi:
Move from nose to pytest
    
nose is no longer maintained, thus we have to move to another
testing system in the long run. This commit updates the make
setup to use pytest instead of nose. However all tests needs
to be migrated to pytest now and no longer use nose/tools


bear454:
Well, I think Travis already approved this commit. Thanks for taking care of that!

PR #109, opened 2016-03-15
Prepare virtual environment for Python

tomschr:
* Add requirement files for development and normal
* Add comment in setup.py's install_requires key (`~=` notation works only for newer pip/setuptools)


schaefi:
very nice thanks

PR #110, opened 2016-03-18
Cleanup storage account commands

schaefi:
listing storage accounts is part of the azurectl storage
command. The compute storage subcommand and its help has been
deleted


bear454:
I would have preferred you do you additions in a separate commit.

Could you at least update the description to note the additions, not just the deletions?

schaefi:
It's just 'azurectl compute storage share help' which was missing, thought we could handle that gracefully

schaefi:
updated the commit message

PR #112, opened 2016-03-24
More setup

bear454:
A few things that got missed in the switch to virtualenv:

* Added some more dependencies to the README
* Updated the git commit hook to use `flake8`


schaefi:
I did not need to specify libffi-devel-gcc5, tested that on a fresh installation (openSUSE 13.2)

PR #114, opened 2016-03-29
Refactor loading of config file

schaefi:
Instead of loading the config file any time a class inherits
from CliTask just provide the load_config capability and
let any task explicitly load the config file when required.

In addition the load_config method creates a new default
config file if no config file existed and the user requested
the automatic configuration file setup on the commandline


bear454:
Nice!

PR #115, opened 2016-03-29
Add a couple small commands for convenience, and make shell more convenient.

bear454:


schaefi:
great, let's take it :+1:

PR #116, opened 2016-03-30
Initial experience use account files not sections

schaefi:
Refactor account setup command
    
Instead of managing several accounts in one config file, each
account is represented by its own config file. Support for
account config files were already in place but the account
setup command did not manage accounts as single files. This
commit changes the behaviour as follows:
    
setup account configure
    
    creates a new account configuration file and does not add
    sections to the existing configuration file
    
setup account remove
    
    removes the account configuration file and not the
    section reference in an existing configuration file
    
Refactor account setup list command
    
Instead of listing only the information for the default account
the command now lists all configuration files and their setup.
In addition an information which config file is used by default
is part of the output
    

schaefi:
ok we are missing to setup the default account. I'm going to add that back

schaefi:
I'm done with this one, are we good to merge ?

azurectl/config.py
```diff
@@ -121,6 +123,33 @@ def get_config_file_list(self):
...
+        if default_exists and not default_islink:
+            raise AzureConfigDefaultLinkError(
+                'Can not link %s as default, %s exists and is not a symlink' %
+                (account_config_file, default_config_file)
+            )
+
+        if default_exists:
+            os.remove(default_config_file)
+
+        os.symlink(account_config_file, default_config_file)
```
bear454: I thought you didn't like symlinking configs? :P

PR #117, opened 2016-04-01
Allow setup of custom Azure SDK in virtualenv

schaefi:
If the pip version of the Azure SDK and the Azure Storage SDK
is not suitable for the development tasks we have to manually
install the code from git into the virtualenv in order to allow
azurectl to make use of this SDK version. This commit adds the
helper script .virtualenv.install-azure-sdk which pulls the
SUSE SDK fork to the local development tree.


.virtualenv.install-azure-sdk
```diff
@@ -0,0 +1,21 @@
...
+    python setup.py build
+    python setup.py install
+)
+branch_name=master
+(
+    cd azure-storage-python
+    git checkout $branch_name
+    python setup.py build
+    python setup.py install
+)
```
rjschwei: Maybe I am missing something here, but it appears to me that we are implying that our changes that need to go upstream are always in "master". I don't think that's a good idea. Rather I would propose that changes that are destined for upstream are in a separate branch, "develop" and thus for travis, assuming that azurectl will want the changes we are pushing for the SDK, we should install form that branch.

schaefi: Hi,

>    Maybe I am missing something here, but it appears to me that we are
>    implying that our changes that need to go upstream are always in
>    "master". I don't think that's a good idea. Rather I would propose that
>    changes that are destined for upstream are in a separate branch,
>    "develop" and thus for travis, assuming that azurectl will want the
>    changes we are pushing for the SDK, we should install form that branch.

For the Azure SDK we made the decision to use the git master branch.
Our fork of the Azure SDK is always in line with the master state we
have used to create the package. That concept is active and working
for quite some time.

If we make a contribution to the Azure SDK which is not (not yet)
merged to master and we make use of it at the same time in azurectl
the branch from which the submit request has been made would have
to be changed in this script. However I don't recommend to work like
this. If we have to change the SDK the process should be:

* form a submit request for the SDK
* wait until merged to master
* update the SDK package
* continue azurectl development with new SDK features

I agree we should not treat the storage SDK that way because they
use pip in the desired way. I'll delete that part

rjschwei: On 04/01/2016 02:10 PM, Marcus Schäfer wrote:
>> +
>> +# Install specified branch
>> +branch_name=master
>> +(
>> +    cd azure-sdk-for-python
>> +    git checkout $branch_name
>> +    python setup.py build
>> +    python setup.py install
>> +)
>> +branch_name=master
>> +(
>> +    cd azure-storage-python
>> +    git checkout $branch_name
>> +    python setup.py build
>> +    python setup.py install
>> +)
> 
> Hi,
> 
>>    Maybe I am missing something here, but it appears to me that we are
>>    implying that our changes that need to go upstream are always in
>>    "master". I don't think that's a good idea. Rather I would propose that
>>    changes that are destined for upstream are in a separate branch,
>>    "develop" and thus for travis, assuming that azurectl will want the
>>    changes we are pushing for the SDK, we should install form that branch.
> 
> For the Azure SDK we made the decision to use the git master branch.
> Our fork of the Azure SDK is always in line with the master state we
> have used to create the package. That concept is active and working
> for quite some time.
> 
> If we make a contribution to the Azure SDK which is not (not yet)
> merged to master and we make use of it at the same time in azurectl
> the branch from which the submit request has been made would have
> to be changed in this script.

That's my point, why write a script that we know has to change when we
are waiting for changes to be accepted upstream?

If all pending changes are always in a branch "devel" then the script
fro travis can simply pull "devel". When there are no pending changes
"devel" == "master". This way there is no churn for the setup script.

> However I don't recommend to work like
> this. If we have to change the SDK the process should be:
> 
> * form a submit request for the SDK
> * wait until merged to master
> * update the SDK package
> * continue azurectl development with new SDK features

I would prefer not to be blocked by SDK changes in the azurectl
development. Since we can keep patches with the package until things get
accepted and released in the upstream SDK we should apply the same
concept to the source.

-- 
rjschwei@suse.com
IRC: robjo

PR #119, opened 2016-04-04
Initial experience allow storage account setup on account configuration

schaefi:
Allow storage/container creation for account setup
    
When configuring a new account setup an optional --create parameter exists which will create the provided storage account and the container as part of the azurectl account configuration
    
### DIFF
--- a/azurectl/setup_account_task.py
+++ b/azurectl/setup_account_task.py
@@ -15,7 +15,7 @@
 usage: azurectl setup account -h | --help
        azurectl setup account configure --name=<account_name> --publish-settings-file=<file>
            [--subscription-id=<subscriptionid>]
-           [--region=<region_name> --storage-account-name=<storagename> --container-name=<containername>]
+           [--region=<region_name> --storage-account-name=<storagename> --container-name=<containername> --create]
        azurectl setup account default --name=<account_name>
        azurectl setup account list
        azurectl setup account region add --region=<region_name> --storage-account-name=<storagename> --container-name=<containername>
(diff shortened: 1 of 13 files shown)


schaefi:
I have not tested if a new sequence of storage account plus container causes a race condition in Azure. Maybe we have to wait for the calls to succeed. Would be nice if we can do a real live test. However I was not sure which account I could use to spawn creation of new storage accounts

Thanks

bear454:
I'll do some live testing...

schaefi:
Thanks

azurectl/setup_account_task.py
```diff
@@ -144,6 +152,44 @@ def __configure_account(self):
         log.info(
             'Added account %s', self.command_args['--name']
         )
+        if self.command_args['--create']:
```
bear454: This is done after the config file is written; in the case where the storage account cannot be created (ex: using an invalid name) - the config then is invalid.

Consider writing the config _after_ all tasks have succeeded, or providing some cleanup mechanism when an interim step fails.

schaefi:
I have added a cleanup step if one of the storage account or container creation steps fails.
However there is no rollback if e.g the storage account was created but the container creation failed for some reason, we will not delete the formerly created storage account.

I would say this is ok, otherwise we need a global transaction concept and rollback which I think would a little bit too much :smile:

schaefi:
I have done many real live tests with account configuration/deletion plus storage account and container setup and I think we are good to merge

PR #120, opened 2016-04-06
Make blobname arg optional on `compute storage upload`

bear454:
It's common to use the source filename as the blob filename, and
Storage#upload already handled that behavior, but it didn't account for how
compressed files are handled automatically.

+ Make the argument optional in the command line
+ Properly adjust the filename of compressed files


schaefi:
nice :+1:

schaefi:
ah right after hitting the merge button I found that the man page was not updated. I'll push that change

bear454:
Oops. Thanks.

PR #121, opened 2016-04-07
Restructure into subpackages

schaefi:
Refactored all sources into sub packages as described in

https://trello.com/c/fvKIVdNP/128-azurectl-refactor-azurectl-into-python-subpackages


schaefi:
successfully tested package build, install and tool

PR #122, opened 2016-04-12
Move compute storage

schaefi:
Move compute storage commands to storage service
    
The subcommands from the compute storage task has been moved
to the storage service. The following new commands exists
    
* storage container
   list, show, create, delete, sas containers
    
* storage share
   list, create, delete file shares
    
* storage disk
   upload, delete, blob data, preferably vhd disks
    
References:

https://trello.com/c/vQ8nMkW3/138-azurectl-move-compute-storage


schaefi:
This is mostly moving code around and changing name spaces. I'm afraid file movement and renaming is always hard to review in patch like results.

rjschwei:
LGTM

PR #126, opened 2016-04-26
Provide custom usage message

schaefi:
Instead of the docopt way to show the usage information we
provide an azurectl specific usage information. The usage
data now always consists of

* the generic call
  ```
  azurectl [global options] service <command> [<args>]
  ```

* the command specific usage defined by the docopt string
  short form by default, long form with -h | --help

* the global options

### DIFF
--- a/azurectl/azurectl.py
+++ b/azurectl/azurectl.py
@@ -17,13 +17,29 @@ import docopt
 # project
 import logger
 from app import App
+from defaults import Defaults
 from azurectl_exceptions import AzureError
+
+
+def extras(help, version, options, doc):
+    """
+    Overwritten method from docopt
+
+    We want to show our own usage for -h|--help
+    """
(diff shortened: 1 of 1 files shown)


schaefi:
I think I have tested all combinations, but the more people try it the better

bear454:
First off, this is a *huge* step in the right direction. Thanks @schaefi :D

One nitpick though, is the double-render on usage, e.g.:
```
usage: azurectl [global options] service <command> [<args>]

usage: azurectl setup account -h | --help
       azurectl setup account configure --name=<account_name> --publish-settings-file=<file>
```

Maybe we avoid that by either adding the `[global options]` piece in the usage lines for the command; it's repetitive, but that's the nature of command definiton.

Thoughts?

schaefi:
I prefer removal one of the usage: statements

PR #129, opened 2016-04-26
Sort commands and options in docopts alphabetically

bear454:


bear454:
I'll sort the entries on the manpages after this is merged.

PR #133, opened 2016-04-27
Compute image usage

schaefi:
Fixed docopt setup for compute image commands

* Options for the image update command were not part of the usage help
* Be more consistent what --name is used for
* Be more clear and consistent for the --label option


bear454:
:+1:

PR #135, opened 2016-04-28
Doc fixes - update manpages & fill in some blanks.

bear454:


schaefi:
great work, thanks for that effort :+1:

Issue #124, opened 2016-04-26
Give the user more information


rjschwei:
We need to tell the user where to find the information they need to set up and account. For example we provide:

azurectl setup account configure \
  --name my_account \
  --management-pem-file /path/to/pem/file \
  --management-url http://url.endpoint \
  --subscription-id 00000000-0000-0000-0000-000000000000
  --region region
  --storage-account-name storage_account_name \
  --container-name container_name

as an example command to set up the configuration for azurectl, but we do not tell the user where to find the information needed. Given that the Azure web console is a freaking mess we should at least make an attempt to provide the user with some guidance as to where to find the the storage account name, container name and other information.

schaefi:
In version 2.0.0 an account can be setup by:

1. publish-settings file
2. management-pem-file, management-url and subscription-id

The first and still preferred method needs the following azurectl call

```
$ azurectl setup account configure --name=bob --publish-settings-file=filename
```

Information where to find or download the publishsettings file is documented here

https://github.com/SUSE/azurectl#publish-settings

Information about regions and storage accounts can be listed right after the account setup with

```
$ azurectl compute vm regions
$ azurectl storage account list
```

The information about containers inside of a storage account requires the name of the storage account. Given you know the name you can list its containers with

```
azurectl --storage-account suseazurerd storage container list
```

With all that information at hand you should be able to add a region setup to the config with

```
azurectl setup account region add --region=<region_name> --storage-account-name=<storagename> --container-name=<containername>
```

All this if you work with publish settings files

schaefi:
If you use the non publish settings based form you need information about

* management-pem-file /path/to/pem/file \
* management-url http://url.endpoint \
* subscription-id 00000000-0000-0000-0000-000000000000

I don't know what else we should write in the documentation other than this is information the Azure provider gives to you when you create an account

But maybe James has an idea how to be more informative here

rjschwei:
Well obviously we know everything that needs to be done. Thus the question remanins, why are we not documenting this, i.e. provide more information to the user, in a more accessible format. Either in the README file or in a separate document that covers setup. I consider the configuration a very critical part of getting people to use the tool. We need to provide explanations about the setup of azurectl in the most detail possible.

schaefi:
I'm fine with extending the README

PR #136, opened 2016-04-29
Not all operations requires a region setup

schaefi:
azurectl operations like 'storage account list' or 'compute vm types'
as well as 'compute vm regions' do not require a region setup in the
configuration file. In addition we allow to setup the account with
the region setup to be optional but the code always reads in a default
region setup. Therefore this commit changes the behaviour to read
the region setup only when required and allows for operations without
a region setup. This is related to Issue #124


rjschwei:
LGTM

PR #141, opened 2016-04-29
Update usage information for global options

### LINKED ISSUE #137: azurctl 2.0 - help message shows too many options
For "azurectl setup account" a number global options are no applicable, thus these should not be displayed, and is provided should be ignored.

For example:

azurectl --account XXX setup account configure --name YYY

makes no sense. Thus when using "azurectl setup account --help" the global option of "--account" should not be displayed. This applies to other global options such as "--config" and "--storage-account" as well. There are others.

IMHO this is a release blocker.

schaefi:
Depending on the selected service different global options
are available. This patch categorizes the global option
information to be specific for the selected service.
Fixes #137

### DIFF
--- a/azurectl/azurectl.py
+++ b/azurectl/azurectl.py
@@ -80,7 +80,7 @@ def usage(command_usage):
     global_options = '\n'
     process_lines = False
     for line in program_code:
-        if line.rstrip() == 'global options:':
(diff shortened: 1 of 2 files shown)


bear454:
:+1:

PR #146, opened 2016-05-01
Vm instance creation

### LINKED ISSUE #144: Can't create a 2nd VM in one cloud service
https://github.com/Azure/azure-sdk-for-python/issues/362

schaefi:
Fixed instance creation if deployment exists
    
check if the virtual machine deployment already exists. If so any new instance must be created as additional role to the deployment. Only the initial virtual machine instance must be created as a new deployment for the selected cloud service Fixes #144

### DIFF
--- a/azurectl/instance/virtual_machine.py
+++ b/azurectl/instance/virtual_machine.py
@@ -152,6 +152,16 @@ class VirtualMachine(object):
                     ' '.join(message) % cloud_service_name
                 )
 
+        if label and deployment_exists:
+            message = [
+                'A deployment of the name: %s already exists.',
+                'Assignment of a label can only happen for the',
+                'initial deployment.'
+            ]
+            raise AzureVmCreateError(
+                ' '.join(message) % cloud_service_name
+            )
+
         media_link = self.storage.make_blob_url(
             self.container_name, ''.join(
                 [
@@ -165,7 +175,6 @@ class VirtualMachine(object):
         instance_record = {
             'deployment_name': cloud_service_name,
             'deployment_slot': group,
-            'label': cloud_service_name,
             'network_config': network_config,
             'role_name': cloud_service_name,
             'role_size': machine_size,
@@ -174,10 +183,6 @@ class VirtualMachine(object):
             'os_virtual_hard_disk': instance_disk,
             'provision_guest_agent': True
         }
-        if reserved_ip_name:
-            instance_record['reserved_ip_name'] = reserved_ip_name
-        if label:
-            instance_record['label'] = label
         if network_config:
             instance_record['network_config'] = network_config
 
@@ -187,6 +192,12 @@ class VirtualMachine(object):
                     **instance_record
                 )
             else:
+                if reserved_ip_name:
+                    instance_record['reserved_ip_name'] = reserved_ip_name
+                if label:
+                    instance_record['label'] = label
+                else:
+                    instance_record['label'] = cloud_service_name
                 result = self.service.create_virtual_machine_deployment(
                     **instance_record
                 )
--- a/azurectl/instance/virtual_machine.py
+++ b/azurectl/instance/virtual_machine.py
@@ -19,6 +19,7 @@ from azure.servicemanagement import OSVirtualHardDisk
 from azure.storage.blob.baseblobservice import BaseBlobService
 
 # project
+from ..management.request_result import RequestResult
 from ..azurectl_exceptions import (
     AzureVmCreateError,
     AzureVmDeleteError,
@@ -132,7 +133,13 @@ class VirtualMachine(object):
         if reserved_ip_name:
             if not deployment_exists:
                 try:
-                    self.service.create_reserved_ip_address(reserved_ip_name)
+                    result = self.service.create_reserved_ip_address(
+                        reserved_ip_name
+                    )
+                    request_result = RequestResult(result.request_id)
+                    request_result.wait_for_request_completion(
+                        self.service
+                    )
                 except Exception as e:
                     raise AzureVmCreateError(
                         '%s: %s' % (type(e).__name__, format(e))
--- a/azurectl/instance/virtual_machine.py
+++ b/azurectl/instance/virtual_machine.py
@@ -122,10 +122,6 @@ class VirtualMachine(object):
                 )
             )
 
-        # check if the virtual machine deployment already exists. If so
-        # any new instance must be created as additional role to the
-        # deployment. Only the initial virtual machine instance must
-        # be created as a new deployment for the selected cloud service
         deployment_exists = self.__get_deployment(
             cloud_service_name
         )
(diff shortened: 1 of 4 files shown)


schaefi:
Please test, I haven't found the time to extensively test this. Thanks

rjschwei:
The code looks OK to me, I did not do any testing.

As a side note I think we need to look some more at the implementation philosophy as reviewing this code led me to believe there are cases when exceptions are used instead of return values. For example the test is a deployment exists or not (service.get_deployment_by_name) throws an exception, when it could simply return the deployment or None. But this is a bigger picture issue and should not influence this particular pull request.

schaefi:
agreed could be done nicer

schaefi:
The code nevertheless needs a real test before we merge, I won't have time to test today, thus if someone has a free slot earlier I would appreciate it

azurectl/instance/virtual_machine.py
```diff
@@ -149,10 +149,28 @@ def create_instance(
...
+        # be created as a new deployment for the selected cloud service
         try:
-            result = self.service.create_virtual_machine_deployment(
-                **instance_record
+            self.service.get_deployment_by_name(
+                service_name=cloud_service_name,
+                deployment_name=cloud_service_name
             )
+            deployment_exists = True
+        except Exception:
```
bear454: This is too general; a network timeout is not the same thing as the deployment not existing; the Exception will need to be interrogated to determine if there was a problem with the API, or if the deployment does not exist.

azurectl/instance/virtual_machine.py
```diff
@@ -149,10 +149,28 @@ def create_instance(
...
+            deployment_exists = False
+
+        try:
+            if deployment_exists:
+                result = self.service.add_role(
+                    **instance_record
```
bear454: `add_role` has a different signature than `create_virtual_machine_deployment`; `instance_record` has to be modified.

azurectl/instance/virtual_machine.py
```diff
@@ -121,6 +121,37 @@ def create_instance(
...
+                deployment_name=cloud_service_name
+            )
+            deployment_exists = True
+        except Exception:
+            deployment_exists = False
+
+        if reserved_ip_name:
+            if not deployment_exists:
+                try:
+                    self.service.create_reserved_ip_address(reserved_ip_name)
```
bear454: This is a breaking change from our existing code; it assumes the `reserved_ip_names` does not already exist. Current workflow requires creating the `reserved_ip` first (`azurectl compute reserved-ip create --name=...`); if a user did that, with this code, it would fail.

While it may be convenient to create a reservation inline, that's not typical for azurectl (for example, we explicitly separate uploading to blob, from registering as image). If you're going to add that convenience here (which IMO isn't necessary) it needs to first check if the `reserved_ip_name` already exists, and is not assigned to a deployment.

azurectl/instance/virtual_machine.py
```diff
@@ -152,6 +152,16 @@ def create_instance(
                     ' '.join(message) % cloud_service_name
                 )
 
+        if label and deployment_exists:
+            message = [
+                'A deployment of the name: %s already exists.',
+                'Assignment of a label can only happen for the',
+                'initial deployment.'
+            ]
```
bear454: This is how `reserved_ip_name` should be handled as well.

bear454:
There are a number of real-world problems here, still. In addition to the existing code, we need to look at how we handle 'role-name', as this will affect how other config options (like creating data-disks) will be able to access the 2nd instance.

azurectl/instance/virtual_machine.py
schaefi: agreed, updated the code

azurectl/instance/virtual_machine.py
schaefi: don't understand that part, I was looking into the sdk code to find the common parts of the two methods. instance_record only contains the elements both methods have in common. Or did I overlook something ?

azurectl/instance/virtual_machine.py
schaefi: you are right, sorry my fault, I'm going to fix that

schaefi:
I was also concerned about the fact that deployment_name, role_name, label and service_name is always set to the cloud_service_name

on the first deployment this seems no problem, however if multiple instances run and role_name always equals cloud_service_name is it then possible to assign a data disk to the right instance ? We have this code here:

```python
result = self.service.add_data_disk(
    cloud_service_name,
    instance_name,
    (role_name or cloud_service_name),
    lun,
    **args
)
```

so what is it based on ? I would say instance_name makes it clear, but am curious what role role_name and deployment_name plays here

azurectl/instance/virtual_machine.py
schaefi: yes, thanks added it

bear454:
What actual data maps back to role_name and deployment_name is not at all clear, and the only way to clarify will be through experimentation. We've made reasonable assumptions, but clearly they don't map to the reality :/

I can see now, though, that what we have doesn't work; we have two VMs in one of our infrastructure's cloud services, both of which were created through the portal (not azurectl) I cannot manage, for example, endpoints to the 2nd VM:

```
$ azurectl compute endpoint list --cloud-service-name=SMT1-Japan-West
{
  "endpoints": [
    {
      "idle-timeout": "4 minutes",
      "instance-port": "22",
      "name": "SSH",
      "port": "2222",
      "protocol": "tcp"
    }
  ]
}

$ azurectl compute endpoint list --cloud-service-name=SMT1-Japan-West --instance-name=SMT1-JP-West-i2
ERROR: AzureEndpointListError: 'AzureMissingResourceHttpError: Not Found\n<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>ResourceNotFound</Code><Message>The deployment name \'SMT1-JP-West-i2\' does not exist.</Message></Error>'
```

bear454:
I'm working through `shell` to try to figure out a good plan...

bear454:
```
>>> svc = service.get_hosted_service_properties('SMT1-Japan-West', embed_detail=True)
>>> len(svc.deployments)
1
>>> [role.role_name for role in svc.deployments[0].role_list]
[u'SMT1-Japan-West', u'SMT1-JP-West-i2']
>>> [instance.instance_name for instance in svc.deployments[0].role_instance_list]
[u'SMT1-Japan-West', u'SMT1-JP-West-i2']
```
So, for instances created through the portal:
* A cloud service is equivalent to a deployment.
* Each virtual machine is created in it's own role

So it looks like we're consistent if we continue to use `deployment_name = cloud_service_name` , but we need to have `role_name = instance_name`.

bear454:
I'll start a separate branch for fixes on instance modules that are affected, if you can just make sure we're creating instances consistently with the portal, here.

bear454:
Instance modules are updated in https://github.com/SUSE/azurectl/pull/151

schaefi:
Thanks much, makes all sense to me. I followed your suggestion here too

bear454:
I'll pull this down and test today. If practical testing looks good, I'll just merge it.

schaefi:
>    I'll pull this down and test today. If practical testing looks good,
>    I'll just merge it.

Thanks for doing this. I'll be off Thursday/Friday, feel free to just
work on the branch if we need additional changes

rjschwei:
LGTM

PR #148, opened 2016-05-01
Make name argument mandatory for container create

schaefi:
If the container name is optional on create, it is taken from
the region section of the config file. However if the config
file does not contain a region section it is not possible to
create a container. This commit solves the chicken & egg
problem. Related to #124


rjschwei:
LGTM

PR #161, opened 2016-05-08
Fixup VirtualMachine and CloudService instances

### LINKED ISSUE #158: VM deletion issue when region is specified on command line but no region in config
When attempting to delete an instance for an account that has no region set in the config file but the region is given on the command line an error is generated:

azurectl --account a_name --region "East US" compute vm delete --cloud-service-name=service_name

ERROR: AzureConfigSectionNotFound: 'Section region:East US not found in configuration file .....

It appears unnecessary that the code tries to consult the configuration for a region if the region is specified on the command line.

schaefi:
The constructor requested information which is not always
needed depending on the called instance methods. Therefore
the class requests operational information like container
names, region name, etc only when needed but not directly
at instance creation time. Fixes #158

### DIFF
--- a/azurectl/instance/cloud_service.py
+++ b/azurectl/instance/cloud_service.py
@@ -35,10 +35,6 @@ class CloudService(object):
     """
     def __init__(self, account):
         self.account = account
-        self.container_name = account.storage_container()
-        self.account_name = account.storage_name()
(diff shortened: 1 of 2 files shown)


rjschwei:
LGTM

PR #162, opened 2016-05-09
Better error message for storage/container config

schaefi:
The information for the storage account and container is connected
to a region section in the configuration file. This means if azurectl
needs to read for example the name of the default storage container
name configured for the default- or commandline provided region,
it has to first read the region section. If the region section
does not exist azurectl provides an error message to the user about
the missing region, however the user requested information about the
storage container and that could confuse users who do not know how
the configuration file is structured. This patch puts together the
complete set of information and forms a more clear error message
for all data read from a region section in the configuration file

### DIFF
--- a/azurectl/config/parser.py
+++ b/azurectl/config/parser.py
@@ -180,16 +180,16 @@ class Config(object):
         return result
 
     def __get_region_option(self, option):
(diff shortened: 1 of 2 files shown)


schaefi:
Related to #158

PR #171, opened 2016-06-13
Better message for default config file conflict

schaefi:
If a default account configuration file from a former
azurectl version was found, an error message explaining
the possible options should be displayed


schaefi:
This was found by Robert during the CSM workshop. My pull request of azurectl was rejected because of the missing storage sdk package in SLE. Therefore we can add this improvement fix for the 2.x release

PS:

The storage sdk package has been submitted

PR #172, opened 2016-07-14
Allow wait for request completion in command

schaefi:
Added --wait for async request, update and fixup the manual pages along with the change

Related to: https://trello.com/c/TDNWTWZW/167-azurectl-next-1

### DIFF
--- a/azurectl/storage/account.py
+++ b/azurectl/storage/account.py
@@ -60,7 +60,7 @@ class StorageAccount(object):
     ):
         try:
             current = self.service.get_storage_account_properties(name)
-            self.service.update_storage_account(
+            result = self.service.update_storage_account(
                 name,
                 (description or current.storage_service_properties.description),
                 (label or current.storage_service_properties.label),
@@ -77,7 +77,7 @@ class StorageAccount(object):
             raise AzureStorageAccountUpdateError(
                 '%s: %s' % (type(e).__name__, format(e))
             )
-        return self.show(name)
+        return result.request_id
 
     def delete(self, name):
         try:
--- a/doc/man/azurectl::compute::data_disk.md
+++ b/doc/man/azurectl::compute::data_disk.md
@@ -14,6 +14,7 @@ __azurectl__ compute data-disk create --cloud-service-name=*name* --size=*disk-s
     [--disk-name=name]
     [--lun=lun]
     [--no-cache|--read-only-cache|--read-write-cache]
+    [--wait]
 
 __azurectl__ compute data-disk list --cloud-service-name=*name*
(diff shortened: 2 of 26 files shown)


schaefi:
I want to have that in and as follow up the wait for an instance to be ready for access before the priority tasks e.g VM reboot. That makes things a little bit easier.

schaefi:
no objections and a pretty simple change set let me think this can be merged in order to unblock the prio task implementations

azurectl/storage/account.py
```diff
@@ -77,7 +77,7 @@ def update(
             raise AzureStorageAccountUpdateError(
                 '%s: %s' % (type(e).__name__, format(e))
             )
-        return self.show(name)
+        return result.request_id
```
bear454: This is an unexpected change in the behavior of the API.

doc/man/azurectl::compute::data_disk.md
```diff
@@ -26,19 +27,17 @@ __azurectl__ compute data-disk show --cloud-service-name=*name* --lun=*lun*
...
+    [--wait]
 
 # DESCRIPTION
 
 ## __create__
 
-Create a new empty disk, and attach it to the selected virtual machine. If the
-virtual machine's __instance-name__ is the same as the __cloud-service-name__,
-the __--instance-name__ argument may be omitted.
+Create a new empty disk, and attach it to the selected virtual machine. If the virtual machine's __instance-name__ is the same as the __cloud-service-name__, the __--instance-name__ argument may be omitted.
```
bear454: I don't consider line breaks at 80 chars to be a "markup issue".

bear454:
There are a couple of issues in this PR that were at least worth discussion, including a change to user output, and dropping the 80-char margin in docs.

Considering that we were all offline Friday, and that your two main reviewers would be travelling over the weekend, you should have waited for a review before merging. Regardless of conditions, we have an agreement to wait 7 days.

azurectl/storage/account.py
schaefi: > > -        return self.show(name)
> > +        return result.request_id
> 
>    This is an unexpected change in the behavior of the API.

The return value was never used in the code, and all async methods
returns the request_id except for this one, which I consider a bug
we overlooked

doc/man/azurectl::compute::data_disk.md
schaefi: >    I don't consider line breaks at 80 chars to be a "markup issue".

yes they are, I'm sorry the rendering of the man pages only works correctly if we don't break the lines ourselves

PR #174, opened 2016-07-21
Cloud service information

schaefi:
Added compute vm show command
    
show detailed information about all virtual machines running in the specified cloud service
https://trello.com/c/PDvqeQZN/168-azurectl-next

### DIFF
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -31,6 +31,7 @@ usage: azurectl compute vm -h | --help
            [--instance-name=<name>]
            [--wait]
        azurectl compute vm regions
+       azurectl compute vm show --cloud-service-name=<name>
        azurectl compute vm types
        azurectl compute vm delete --cloud-service-name=<name>
            [--instance-name=<name>]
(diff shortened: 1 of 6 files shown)


schaefi:
@bear454 no idea why it complains about code duplication in the test, I also would prefer things like landscape, codacy and friends to provide as report and statistics but not cause a fail result. For that purpose we have the unit tests, the coverage and the flake8 checks, that imho should be enough

azurectl/commands/compute_vm.py
```diff
@@ -31,6 +31,7 @@
            [--instance-name=<name>]
            [--wait]
        azurectl compute vm regions
+       azurectl compute vm show --cloud-service-name=<name>
```
rjschwei: Does it makes sense, or is there even a way to provide info about a specific VM?

azurectl compute vm show --cloud-service-name

provides info about all the VMs in the cloud service and more and

azurectl compute vm show --cloud-service-name=... -vm-name=...

would provide information only about this specific machine within the cloud service?

schaefi: The sdk method allows only to set details=True and that will then provide the complete information. We could however think of reducing the information to the specified instance name

schaefi: However I don't see it as that important

bear454:
@schaefi: sorry, I didn't realize I had set up codacy to comment here; I'll try to turn those off for now.

bear454:
Looks good to me.

PR #175, opened 2016-07-22
Fixed OS image lookup in VirtualMachine

schaefi:
Before we start an instance a check is made if the provided image
is reachable by the cloud service. However if the image itself
does not (or not yet) exist azurectl runs into an unhandled
exception

### DIFF
--- a/azurectl/instance/virtual_machine.py
+++ b/azurectl/instance/virtual_machine.py
@@ -252,8 +252,12 @@ class VirtualMachine(object):
         ).storage_service_properties.location
 
     def __image_locations(self, disk_name):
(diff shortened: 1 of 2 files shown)


bear454:
We should explain to the user at some point *why* his request failed.

schaefi:
>    We should explain to the user at some point why his request failed.

With this fix the AzureImageNotReachableByCloudServiceError is
raised causing this message

```
'The selected image "%s" is not available',
'in the region of the selected cloud service "%s",',
'cannot launch instance'
```

I think that matches that case and is clear enough

Thoughts ?

PR #176, opened 2016-07-26
Retain data disk

schaefi:
Allow to retain associated data disk
    
The option --delete-vhd=True|False allows to retain an associated data disk when deleting it from the instance. By default the data disk won't be destroyed, which is a change in the default behaviour of the command

### DIFF
--- a/azurectl/commands/compute_data-disk.py
+++ b/azurectl/commands/compute_data-disk.py
@@ -16,39 +16,57 @@ Data-disks are virtual disks attached to a virtual machine, backed by a virtual
 hard disk (VHD) image in Azure storage.
 
 usage: azurectl compute data-disk -h | --help
-       azurectl compute data-disk create --cloud-service-name=<name> --size=<disk-size-in-GB>
+       azurectl compute data-disk create --identifier=<name> --size=<disk-size-in-GB>
+       azurectl compute data-disk delete --disk-name=<name>
+       azurectl compute data-disk attach --cloud-service-name=<name> --disk-name=<name>
            [--instance-name=<name>]
            [--label=<label>]
-           [--disk-name=<name>]
            [--lun=<lun>]
            [--no-cache|--read-only-cache|--read-write-cache]
            [--wait]
-       azurectl compute data-disk list --cloud-service-name=<name>
-           [--instance-name=<name>]
-       azurectl compute data-disk show --cloud-service-name=<name> --lun=<lun>
-           [--instance-name=<name>]
-       azurectl compute data-disk delete --cloud-service-name=<name> --lun=<lun>
+       azurectl compute data-disk detach --cloud-service-name=<name> --lun=<lun>
            [--instance-name=<name>]
            [--wait]
+       azurectl compute data-disk list
+       azurectl compute data-disk list attached --cloud-service-name=<name>
+           [--instance-name=<name>]
+       azurectl compute data-disk show --disk-name=<name>
+       azurectl compute data-disk show attached --cloud-service-name=<name> --lun=<lun>
+           [--instance-name=<name>]
        azurectl compute data-disk help
 
 commands:
     create
-        add a new, empty data disk to the selected virtual machine
+        create a new, empty data disk. The data disk vhd file will be
+        created using the following naming schema:
+        <identifier>-data-disk-<utctime>
     delete
-        detach a data disk from the selected virtual machine and destroy the
-        data file
+        delete the specified data disk. The call will fail if the disk
+        is still attached to an instance
+    attach
+        attach the specified data disk vhd file to the selected virtual machine
+    detach
+        detach a data disk from the selected virtual machine and retain the
+        data disk vhd file
     list
-        return data about all data disks attached to a virtual machine
+        return list of all disks from your image repository
+    list attached
+        return list of all data disks attached to a virtual machine
     show
-        return data about an existing data disk
+        return details of the specified disk
+    show attached
+        return details about the data disk attached to the selected
+        virtual machine and lun
 
 options:
     --cloud-service-name=<name>
         name of the cloud service where the virtual machine may be found
     --disk-name=<name>
-        name of the disk file created in the current storage container. If
-        omitted, a name will be automatically generated.
+        name of the data disk created in the current storage container.
+    --identifier=<name>
+        identifier string used as part of the complete data disk name.
+        Usually this is set to the instance name this data disk should be
+        attached to later.
     --instance-name=<name>
         name of the virtual machine instance. If no name is given the
         instance name is assumed to be the same as the cloud service name
@@ -72,6 +90,7 @@ options:
         wait for the request to succeed
 """
 # project
+from ..logger import log
 from ..account.service import AzureAccount
 from base import CliTask
 from ..utils.collector import DataCollector
@@ -101,22 +120,25 @@ class ComputeDataDiskTask(CliTask):
 
         self.account = AzureAccount(self.config)
         self.data_disk = DataDisk(self.account)
-        self.data_disk.set_instance(
-            self.command_args['--cloud-service-name'],
-            (
-                self.command_args['--instance-name'] or
-                self.command_args['--cloud-service-name']
-            )
-        )
 
         if self.command_args['create']:
             self.__create()
-        if self.command_args['show']:
-            self.__show()
         if self.command_args['delete']:
             self.__delete()
-        if self.command_args['list']:
-            self.__list()
+        if self.command_args['attach']:
+            self.__attach()
+        if self.command_args['detach']:
+            self.__detach()
+        if self.command_args['attached']:
+            if self.command_args['list']:
+                self.__list_attached()
+            if self.command_args['show']:
+                self.__show_attached()
+        else:
+            if self.command_args['list']:
+                self.__list()
+            if self.command_args['show']:
+                self.__show()
 
     def __help(self):
         if self.command_args['help']:
@@ -126,11 +148,20 @@ class ComputeDataDiskTask(CliTask):
         return self.manual
 
     def __create(self):
+        self.data_disk.create(
+            self.command_args['--identifier'],
+            self.command_args['--size'],
+            self.command_args['--label']
+        )
+        log.info(
+            'Created new data disk %s',
+            self.data_disk.data_disk_name.replace('.vhd', '')
+        )
+
+    def __attach(self):
         optional_args = {}
         if self.command_args['--label']:
             optional_args['label'] = self.command_args['--label']
-        if self.command_args['--disk-name']:
-            optional_args['filename'] = self.command_args['--disk-name']
         if self.command_args['--lun']:
             optional_args['lun'] = int(self.command_args['--lun'])
         if (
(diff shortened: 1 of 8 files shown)


doc/man/azurectl::compute::data_disk.md
```diff
@@ -37,7 +38,7 @@ Create a new empty disk, and attach it to the selected virtual machine. If the v
 
 ## __delete__
 
-Detach the data disk from the selected __lun__ of the selected virtual machine and destroy the data file.
+Detach the data disk from the selected __lun__ of the selected virtual machine. The data disk is destroyed if __--delete-disk__ is provided with the command. By default the data disk is not destroyed.
```
bear454: I think it would more sense to have a separate command here (maybe `compute data-disk detach` ?); having the `delete` command *not delete by default* feels incongruous.

rjschwei:
We have some conceptual inconsistencies here.

We should have commands create, delete, attach, detach.

- I should be able to create a disk without attaching it to a running instance immediately
- I should be able to delete a created disk, only if it is not attached of course ;)
- I should be able to attach any disk I have laying around (in the same region) to any instance
- I should be able to detach a disk from any instance

This is a behavioral change for the existing create and delete commands but makes sense conceptually and matches with behavior in the physical world. I can buy a hard drive that is not part of a computer and then I ca plug that hard drive into any computer I please.

schaefi:
yep makes all sense, I'm going to rework the code

schaefi:
needless to say that fixing the conceptional problems you correctly outlined had a big impact on the code. Thus this planed simple and compatible change has grown into a bigger refactoring. The data_disk command also has changed its behaviour which means this will also cause a major version update

I have not completed real live testing, any help is appreciated

schaefi:
no testing needed, I need to rework that again

schaefi:
I'm afraid the Azure api has no easy way to create a data disk without attaching it to an instance. The
DataDisks api is always connected to an instance. I tried a different way with:

* create disk blob
* call add_disk
* call add_data_disk with existing disk

The problem here is that the disk blob must be a valid vhd. There is no api for creating a vhd you can only create a blob. In order to make the blob a vhd we have to upload a vhd signature

It seems the DataDisks api internally creates a vhd but that functionality is not populated to the public

I'm still coding a solution

schaefi:
Ha, found a solution to the problem and tested data disk creation/deletion attach/detach live in Azure

will come up with a revised pull request soon

schaefi:
All right I'm finished and the data-disk command now works as suggested by Robert. I have tested it but it would be good if others do some tests too.

I'm afraid the code has changed in several places and the review might be harder than the original issue pointed out

azurectl/commands/compute_data-disk.py
```diff
@@ -16,39 +16,57 @@
 hard disk (VHD) image in Azure storage.
 
 usage: azurectl compute data-disk -h | --help
-       azurectl compute data-disk create --cloud-service-name=<name> --size=<disk-size-in-GB>
+       azurectl compute data-disk create --identifier=<name> --size=<disk-size-in-GB>
+       azurectl compute data-disk delete --disk-name=<name>
+       azurectl compute data-disk attach --cloud-service-name=<name> --disk-name=<name>
            [--instance-name=<name>]
            [--label=<label>]
-           [--disk-name=<name>]
```
bear454: Why is this mandatory now?

bear454:
@rjschwei said:
> I should be able to create a disk without attaching it to a running instance immediately

We already have that functionality for a populated disk: `storage disk upload`. I don't see any value in creating a blank, empty disk that's not attached, especially since we know from the current implementation that the empty disk can be created on-the-fly when it's attached.

The implementation presented here, which uses `qemu-img` to create a blank disk, then uploads it just seems like a waste of resources, and adds qemu as a required dependency of `azurectl`. :/

azurectl/instance/data_disk.py
```diff
@@ -32,97 +36,187 @@ class DataDisk(object):
...
-            lun = self.__get_first_available_lun()
+        """
+            Attach existing data disk to the instance
+        """
+        disk_url = self.__data_disk_url(disk_name + '.vhd')
+
+        if not instance_name:
+            instance_name = cloud_service_name
+
+        if lun not in range(Defaults.max_vm_luns()):
```
bear454: :+1:

azurectl/instance/data_disk.py
```diff
@@ -32,97 +36,187 @@ class DataDisk(object):
...
+        self.attached_lun = None
+
+    def create(self, identifier, disk_size_in_gb, label=None):
+        """
+            Create new data disk
+        """
+        disk_vhd = NamedTemporaryFile()
+        vhd_image = subprocess.Popen(
+            [
+                'qemu-img', 'create', '-f', 'vpc', '-o',
```
bear454: As I mentioned in general comments; I'd rather we leave this out... I'd be fine with including some additional documentation about localized ways of generating a blank disk (if there's any valid use case for that) - but making `qemu` a dependency (think of the mess that is 'qemu for windows') of *azurectl* seems like a bad approach.

rjschwei: I agree that a dependency on qemu is undesirable. We aim to have azurectl be a platform independent tool and forcing qemu onto a Windows system in order to use azurectl creates a problem since it is not a simple pip install away or provided by Msft in any way.

rjschwei:
@bear454 

Why is creation of an empty disk not useful? Comparing it to the upload of a populated disk does not appear to be a good analogy to me. If I want an empty disk I don't want to upload a diskimage that was created with kiwi or other tool.

We also need to keep the implementation separate from the conceptual discussion. While I agree that a dependency on qemu is not desirable that has nothing to do with the concept of creating an empty disk.

From my point of view having to create an instance to create an empty disk appears to be conceptually broken. Like saying that if I want a physical hard drive I have to buy a computer.

Back to the conceptual behavior I stand by my earlier comment of things being symmetric:

- create <-> delete
- attach <-> detach

Maybe based on functionality of the platform this symmetric behavior needs to have some restrictions.

rjschwei:
Now to the implementation, purposefully using a new comment to separate the concept of symmetric use from implementation detail of the concept.

- create
  + If creation of an empty disk is not possible via API without it being attached to an instance we have a number of options:
    1.) Work with Microsoft to see if there is an API that is simply not exposed in the SDK,
          a.) If the API exists we can contribute to the SDK
          b.) If the API does not exist my guess would be that it will not be implemented in ASM thus this
               would be a dead end
    2.) We can stick with the proposed solution here, but this comes at the heavy price of a qemu
          dependency which is IMHO undesirable
    3.) We have the users information, we can under the cover fire up a helper instance, create
          an attached empty disk, detach and kill the instance
    4.) We can tie the create function to require a running instance. Going down this road would have to
          account for the fact that a cloud service may contain multiple VMs and thus we need a way to
          sort out to which instance the empty disk device would be attached.
    5.) We could have a mix between 3&4 by adding an option such as --auto-instance that would 
          implement the behavior of 3 and if not set the behavior of 4 would be enforced
    ... There may be other options

- delete, attach, and detach
  + Appear to be reasonably straight forward

schaefi:
We should do a better job in defining what we want

* The Azure API allows to create a data disk only in combination
  with a running instance

* I agree to what Robert said that azurectl should be able to
  create a data disk _without_ the need to attach it immediately.
  This allows for a way better workflow of create / attach / detach
  as explained by Robert

* In order to achieve the above a way to create and register an
  empty vhd as data disk to Azure must be implemented. This implies
  the need to create a vhd formatted sparse file of the correct
  vhd-fixed format. On Linux this can only be done with qemu.
  The Azure API has no call to do this. yes the consequence is the
  tool is Linux agnostic for this particular use case. I personally
  don't care because it can be easily documented how a Windows user
  can create a VHD and upload it. If you google search this you will
  even find instructions directly targeted to Azure, none for Linux
  though. I'm open to improve the exception message to inform the
  user better.

* If we don't want an instance independent data disk registration
  with azurectl we can go back to the original implementation.

I have tested data disk usage with the suma appliance and I really
like the way it works now. I find it pretty inconvenient that I need
an instance before I can attach a data disk to it.

The creation of a correctly sized, correctly named, and correctly
formatted data disk in Azure requires the user to know about
several details. It should not be the task of the user to upload
its own hand crafted vhd to register a data disk with it

my 2cents

bear454:
First, I don't see any value in having a blank disk in storage that is not attached to any machine, and that is how Azure treats data disks as well:

From [the docs](https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-windows-classic-attach-disk/)
> Attaching an empty disk is a simple way to add a data disk, because Azure creates the .vhd file for you and stores it in the storage account.

... which is roughly based on the definition of the ['Add Data Disk' REST command](https://msdn.microsoft.com/en-us/library/azure/jj157199.aspx).

Second, I think it's out of scope to create a disk image in `azurectl`. We have a method to upload a VHD file locally to a page blob in Azure; whether that VHD is populated or completely blank is really moot.

That being said, both of you seem to think there is value. I did some research, and found a [blog post](http://webcache.googleusercontent.com/search?q=cache:pk-rGqnXQQYJ:blog.stevenedouard.com/create-a-blank-azure-vm-disk-vhd-without-attaching-it/+&cd=6&hl=en&ct=clnk&gl=us) from someone trying to accomplish the same thing. Simply put, we don't need to generate the whole disk image, just the fixed-size VHD footer, which can be done in pure python:

https://gist.github.com/sedouard/0468d1b2d0153ca9a652#file-blank_vhd-py

So, can we request an open license on the gist, and embed this code, or something derivative, instead of creating a disk image file?

rjschwei:
Well there is a reference to a spec in the code. Thus we should be able to come up with code that does the same thing and is not a copy thereof.

https://technet.microsoft.com/en-us/virtualization/bb676673.aspx#E3B

rjschwei:
So one use case of having an empty disk laying around is to attach it to different instances at different times for different purposes. One could use the disk for information sharing, for example. If one has a system that is completely disconnected from the outside world one could put packages on the disk that are downloaded and verified and tested on one system that is attached to the outside world. Then detach the previously empty disk and attach it to the instance that has not outside world network connection.

Now this of course only illustrates that having a blank disk at hand is convenient, it does not necessarily make the argument that one should be able to create such an empty disk without it originally being attached to an instance. However, it appears useful to be able to create such a disk in a state that does not force the attachment. But we could, as provided as an option earlier always force the attachment and then after detaching one is in the same state. Although I prefer the creation in a non attached state. But granted that is a personal preference.

schaefi:
I agree 100% to what Robert said

All of these discussions however do not belong into a pull request open for reviewing an implementation to fix an issue I thought we had planned and discussed. Obviously we first have to come to an agreement of the overall behaviour of the data-disk command.

Therefore I'm closing this now and updated the trello card task list to discuss the issues first

Overall I'm afraid to say that this is exactly the result of bad planning, especially think of the developers coding work and the enthusiastic being destroyed in a second just because not all people of the development team agree with the concept. There are some fundamental questions where we seem to have a different opinion. I've added those to the card.

I also suggest that we double check as a team if the concept of a command is still ok for us before we add any additional tasks for it in the future

I'll keep another branch with the changes because it technically fixes both open tasks with the implemented attach and detach subcommands of the data-disk command. Maybe the work was not completely for nothing

rjschwei:
On 08/02/2016 03:47 AM, Marcus Schäfer wrote:
> I agree 100% to what Robert said
> 
> All of these discussions however do not belong into a pull request open for reviewing an implementation to fix an issue I thought we had planned and discussed. Obviously we first have to come to an agreement of the overall behaviour of the data-disk command.
> 
> Therefore I'm closing this now and updated the trello card task list to discuss the issues first
> 
> Overall I'm afraid to say that this is exactly the result of bad planning, especially think of the developers coding work and the enthusiastic being destroyed in a second just because not all people of the development team agree with the concept. There are some fundamental questions where we seem to have a different opinion. I've added those to the card.
> 
> I also suggest that we double check as a team if the concept of a command is still ok for us before we add any additional tasks for it in the future
> 
> I'll keep another branch with the changes because it technically fixes both open tasks with the implemented attach and detach subcommands of the data-disk command. Maybe the work was not completely for nothing

The work was definitely not for nothing. Sometimes conceptual issues
only surface after there is an initial implementation to "touch, feel,
and see". This is what happened here.

bear454:
@schaefi said
> Overall I'm afraid to say that this is exactly the result of bad planning, especially think of the developers coding work and the enthusiastic being destroyed in a second just because not all people of the development team agree with the concept. There are some fundamental questions where we seem to have a different opinion. I've added those to the card.

I disagree. The initial implementation had some fundamental issues, with a `delete` command that *didn't delete by default*, so there was rework to be done in any case, and you agreed:

> yep makes all sense, I'm going to rework the code

If there was any fundamental mistake, is was in growing the scope of the PR to try to add _all_ the `data-disk` functions, instead of staying in the scope of separating `delete` from `detach`. Beyond that, I think we had some valuable discussion about how to change `create` and `attach`,  but the name of this PR is still **Retain data disk**.

PR #178, opened 2016-08-10
Show error details from docopt if present

schaefi:
If there are error details from a wrong formatted command line show them at the end of the usage output


schaefi:
The problem here is that docopt does not always populate error details.

Example:

```
azurectl --account

usage...

--account requires argument
```

:+1:  but

```
azurectl --account compute vm types

usage...

<no error details available>
```

:-1:

rjschwei:
Well lets see how this goes, I think we might have to do a little more probing on our own in the long run, but it's a start.

PR #181, opened 2016-08-10
Data disk detach delete

schaefi:
* Refactor DataDisk class
* Update compute data-disk command

### DIFF
--- a/azurectl/commands/compute_data-disk.py
+++ b/azurectl/commands/compute_data-disk.py
@@ -23,32 +23,42 @@ usage: azurectl compute data-disk -h | --help
            [--lun=<lun>]
            [--no-cache|--read-only-cache|--read-write-cache]
            [--wait]
-       azurectl compute data-disk list --cloud-service-name=<name>
-           [--instance-name=<name>]
-       azurectl compute data-disk show --cloud-service-name=<name> --lun=<lun>
-           [--instance-name=<name>]
-       azurectl compute data-disk delete --cloud-service-name=<name> --lun=<lun>
+       azurectl compute data-disk delete --disk-name=<name>
+       azurectl compute data-disk detach --cloud-service-name=<name> --lun=<lun>
            [--instance-name=<name>]
            [--wait]
+       azurectl compute data-disk list
+       azurectl compute data-disk show --disk-name=<name>
+       azurectl compute data-disk show attached --cloud-service-name=<name>
+           [--lun=<lun>]
+           [--instance-name=<name>]
        azurectl compute data-disk help
 
 commands:
     create
-        add a new, empty data disk to the selected virtual machine
+        create a new, empty data disk attached to the specified instance.
+        The data disk vhd file will be created using the following naming
+        schema: <instance-name|cloud-service-name>-data-disk-<utctime>
     delete
-        detach a data disk from the selected virtual machine and destroy the
-        data file
+        delete the specified data disk. The call will fail if the disk
+        is still attached to an instance
+    detach
+        detach a data disk from the selected virtual machine and retain the
+        data disk vhd file
     list
-        return data about all data disks attached to a virtual machine
+        return list of all disks from your image repository
     show
-        return data about an existing data disk
+        return details of the specified disk
+    show attached
+        return details about the data disk(s) attached to the selected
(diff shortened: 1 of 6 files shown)


schaefi:
According to our team meeting this one adds the delete/detach capability, next would be create/attach

schaefi:
I'll kept the revised list and show methods from my former pull request since we are moving this command to work with attached and not-attached disks

bear454:
The implementation looks fine, @schaefi (although you did do a little feature blow with the changes to `data-disk list` and `data-disk show`).

I'm looking at the implementation in the API, though, and I think in the end we're creating duplicate functionality...

the `data-disk detach` command is calling `ServiceManagementService#delete_data_disk()`, but retaining the VHD file; the page blob, in effect all 'data-disk' metadata is destroyed at that point, as the 'data-disk' metadata binds the backing page blob to the VM. (Correct me if I'm wrong, please). So, at this point, is `data-disk delete` any different than `storage disk delete` ?

schaefi:
The call of ```ServiceManagementService#delete_data_disk``` and retaining the vhd does not delete the created disk metadata in my tests. I have tested this as follows:

```
azurectl compute data-disk create \
    --cloud-service-name suse-products --instance-name suma --size 8
```

I got

* the vhd at https://susesuselinuxenterprises.blob.core.windows.net/vm-images/suma-data-disk-2016-08-12T07_59_17.493488.vhd
* the disk reference with the disk-name: suse-products-suma-0-201608120759180096

the disk is attached as shown in

```
azurectl compute data-disk list

{
    "is_attached": true,
    "name": "suse-products-suma-0-201608120759180096"
}
```

Now if I detach the disk with

```
azurectl compute data-disk detach \
    --cloud-service-name suse-products --instance-name suma --lun 0
```

the disk is no longer available to the instance but it still exists and the above __is_attached__ flag changes to ```false```

At that point I can't delete the blob storage because the disk reference is still present.
Therefore I call

```
azurectl compute data-disk delete \
    --disk-name suse-products-suma-0-201608120759180096
```

and that removes the disk and the vhd

I don't think we are duplicating something, at least to the degree I understand the concept with disks in Azure

Thoughts ?

rjschwei:
So there appears to be some remnant of data that prevents "storage disk delete" from working the question is, is this a timing issue? Meaning can you re-run you test but lets say wait an hour after the "data-disk detach" call to see if "storage disk delete" will work then?

Not that I am advocating that we should make users wait for any significant amount of time, it would just be nice to figure out if this is a timing issue or not.

schaefi:
>    So there appears to be some remnant of data that prevents "storage disk
>    delete" from working the question is, is this a timing issue? Meaning
>    can you re-run you test but lets say wait an hour after the "data-disk
>    detach" call to see if "storage disk delete" will work then?

Hmm, good point. All I can say is that after detach it just takes seconds
and the disk is gone from the instance and persists in Azure with the
disk metadata flag __attached_to__ set to None. I think it would be really
weird to maintain such a flag in the metadata of the disk if this device is
about to disappear non deterministicly after whatever amount of time

But I'll give it a try tomorrow and will run the detach and leave it that
way for a day

schaefi:
ok test running, check yourself

```
azurectl --account publish compute data-disk list

{
      "is_attached": false,
      "name": "suse-products-suma-0-201608160936170839"
}

azurectl --account publish compute data-disk show --disk-name suse-products-suma-0-201608160936170839

{
  "data-disk": {
    "affinity_group": "",
    "attached_to": {},
    "has_operating_system": "",
    "is_corrupted": "",
    "label": "",
    "location": "West US",
    "logical_disk_size_in_gb": "8 GB",
    "media_link": "https://susesuselinuxenterprises.blob.core.windows.net/vm-images/suma-data-disk-2016-08-16T09_36_17.008765.vhd",
    "name": "suse-products-suma-0-201608160936170839",
    "os": "",
    "source_image_name": ""
  }
}
```

I don't expect this one to go away

I'll keep this open for a day starting now

bear454:
@schaefi thanks for diving into it; appreciated. I'm good with this implementation, but please be mindful of `storage disk` commands as you move forward.

schaefi:
will do, thanks :+1:

Issue #177, opened 2016-08-05
Custom data from a file when creating a VM


bear454:
`azurectl compute vm create` allows injection of custom data, but as a base64-encoded string. There should be an option to specify a file path, then open & encode the file, and pass it along.

bear454:
Assigned to myself.

PR #182, opened 2016-08-15
Handle custom data via file or string on vm creation

bear454:
re: https://github.com/SUSE/azurectl/issues/177

Also resolves an issue with double-encoding of customdata - the custom
data supplied _should not_ be base64-encoded; this is handled by the
API.

### DIFF
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -17,7 +17,7 @@ Virtual machines are created in a private IP address space, and attached to a
 
 usage: azurectl compute vm -h | --help
        azurectl compute vm create --cloud-service-name=<name> --image-name=<image>
-           [--custom-data=<string> | --custom-data-file=<file>]
+           [--custom-data=<string-or-file>]
            [--instance-name=<name>]
            [--instance-type=<type>]
            [--label=<label>]
@@ -62,10 +62,8 @@ options:
     --cloud-service-name=<name>
         name of the cloud service to put the virtual machine in.
         if the cloud service does not exist it will be created
-    --custom-data=<string>
-        a string of data that will be injected into the new virtual machine
-    --custom-data-file=<file>
-        path to a file, from which the contents will be injected into the new
+    --custom-data=<string-or-file>
+        a string of data or path to a file that will be injected into the new
         virtual machine
     --fingerprint=<thumbprint>
         thumbprint of an already existing certificate in the
@@ -98,6 +96,7 @@ options:
     --wait
         wait for the request to succeed
 """
+import os
 # project
 from base import CliTask
 from ..account.service import AzureAccount
@@ -271,8 +270,8 @@ class ComputeVmTask(CliTask):
         password = self.command_args['--password']
 
         custom_data = self.command_args['--custom-data']
-        if self.command_args['--custom-data-file']:
-            with open(self.command_args['--custom-data-file'], 'r') as file:
+        if (custom_data and os.path.isfile(custom_data)):
+            with open(custom_data, 'r') as file:
                 custom_data = file.read()
 
         disable_ssh_password_authentication = False
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -17,7 +17,7 @@ Virtual machines are created in a private IP address space, and attached to a
 
 usage: azurectl compute vm -h | --help
        azurectl compute vm create --cloud-service-name=<name> --image-name=<image>
-           [--custom-data=<base64_string> | --custom-data-file=<file>]
+           [--custom-data=<string> | --custom-data-file=<file>]
            [--instance-name=<name>]
            [--instance-type=<type>]
            [--label=<label>]
@@ -62,12 +62,11 @@ options:
     --cloud-service-name=<name>
         name of the cloud service to put the virtual machine in.
         if the cloud service does not exist it will be created
-    --custom-data=<base64_string>
+    --custom-data=<string>
         a string of data that will be injected into the new virtual machine
-        after being base64-encoded.
     --custom-data-file=<file>
         path to a file, from which the contents will be injected into the new
-        virtual machine, after being base64-encoded.
+        virtual machine
     --fingerprint=<thumbprint>
         thumbprint of an already existing certificate in the
         cloud service used for ssh public key authentication
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -17,7 +17,7 @@ Virtual machines are created in a private IP address space, and attached to a
 
 usage: azurectl compute vm -h | --help
        azurectl compute vm create --cloud-service-name=<name> --image-name=<image>
-           [--custom-data=<base64_string>]
+           [--custom-data=<base64_string> | --custom-data-file=<file>]
            [--instance-name=<name>]
            [--instance-type=<type>]
            [--label=<label>]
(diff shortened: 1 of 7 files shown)


bear454:
Reference source: https://azure.microsoft.com/en-us/blog/custom-data-and-cloud-init-on-windows-azure/

azurectl/commands/compute_vm.py
```diff
@@ -17,7 +17,7 @@
 
 usage: azurectl compute vm -h | --help
        azurectl compute vm create --cloud-service-name=<name> --image-name=<image>
-           [--custom-data=<base64_string>]
+           [--custom-data=<base64_string> | --custom-data-file=<file>]
```
rjschwei: Lets just figure out if the user gives us a string or a file without adding another argument.

azurectl/commands/compute_vm.py
```diff
@@ -63,8 +63,11 @@
         name of the cloud service to put the virtual machine in.
         if the cloud service does not exist it will be created
     --custom-data=<base64_string>
-        base64 encoded data string. The information is available
-        from the walinux agent in the running virtual machine
+        a string of data that will be injected into the new virtual machine
+        after being base64-encoded.
```
rjschwei: The new description implies that the string is "plain text" yet the custom-data argument still implies the argument should be base64 encoded, which one is it?

azurectl/commands/compute_vm.py
```diff
@@ -267,7 +270,12 @@ def __prepare_linux_configuration(self, fingerprint=u''):
         user = self.command_args['--user']
         instance_name = self.command_args['--instance-name']
         password = self.command_args['--password']
+
         custom_data = self.command_args['--custom-data']
+        if self.command_args['--custom-data-file']:
```
rjschwei: So if we get rid of custom-data-file this would be:

if os.path.isfile(custom_data):
   ......

doc/man/azurectl::compute::vm.md
```diff
@@ -71,9 +71,15 @@ List available instance types and their attributes
 
 Name of the cloud service to put the virtual machine in. If the cloud service does not exist it will be created.
 
-## __--custom-data=base64_string__
+## __--custom-data=string__
 
-A base64 encoded raw stream. The information is available from the walinux agent in the running virtual machine.
+A string of data that will be into injected the new virtual machine after being base64-encoded. Waagent will store this base64-encoded data on the VM as both an attribute of __/var/lib/waagent/ovf-env.xml__ and as the sole contents of __/var/lib/waagent/CustomData__.
```
rjschwei: injected into

azurectl/instance/virtual_machine.py
```diff
@@ -222,6 +224,14 @@ def reboot_instance(
                 '%s: %s' % (type(e).__name__, format(e))
             )
 
+    def validate_custom_data_length(self, custom_data):
```
schaefi: should this be part of the public interface of the class, or is it rather a private method ?

azurectl/commands/compute_vm.py
bear454: I'm fine with making that change, but please be aware it's inconsistent with our other args: e.g. `[--ssh-private-key-file=<file> | --fingerprint=<thumbprint>]`

azurectl/commands/compute_vm.py
bear454: The API performs a base64 encode, and the agent writes the base64-encoded data to disk. Inputs should all be 'plain'.

bear454: If base64-encoded data is provided as input, it will need to be decoded *twice*.

azurectl/commands/compute_vm.py
bear454: :+1:

doc/man/azurectl::compute::vm.md
bear454: thanks

azurectl/commands/compute_vm.py
rjschwei: Sorry I don;t see the relationship here.  The fingerprint has the purpose of verifying a key while ssh-private-key-file provides the key itself inside of a file.

custom-data and custom-data-file both provide the custom data that is supposed to be injected.

azurectl/commands/compute_vm.py
rjschwei: OK, so obviously I did not explain the cross-reference properly. In the help that would be generated it would read:

--custom-data=<base64_string>  (see line 20)

which implies an encoded string is expected. However the long text says it should be plain text. This is a contradiction, at least the way I interpret the code. Thus it should read:

--custom-data=<string>

on line 20.

azurectl/instance/virtual_machine.py
bear454: It can be private; I just don't really like all the underscores.

schaefi: agreed, I also like more to have a way to say something is private. However, if you make it part of the public interface it means people are allowed to make use of it when they import our classes. If that's fine, and I don't have a strong opinion with this particular method, it's all ok.

azurectl/commands/compute_vm.py
bear454: Resolved in f4dec9b.

azurectl/commands/compute_vm.py
bear454: Updated in d86e1c5.

doc/man/azurectl::compute::vm.md
bear454: Fixed in 8bbe556.

azurectl/instance/virtual_machine.py
bear454: Resolved in 9f4b27a

doc/man/azurectl::compute::vm.md
```diff
@@ -71,9 +71,11 @@ List available instance types and their attributes
 
 Name of the cloud service to put the virtual machine in. If the cloud service does not exist it will be created.
 
-## __--custom-data=base64_string__
+## __--custom-data=string-or-file__
 
-A base64 encoded raw stream. The information is available from the walinux agent in the running virtual machine.
+A string of data, or a path to a file whose contents will be injected into the new virtual machine after being base64-encoded. Waagent will store this base64-encoded data on the VM as both an attribute of __/var/lib/waagent/ovf-env.xml__ and as the sole contents of __/var/lib/waagent/CustomData__.
+
+Note: customdata is limited to 64K; using a file larger than 64K will fail.
```
schaefi: I think we could add the information how the data is stored on the machine is depending on the waagent setup in Provisioning.DecodeCustomData and outside of our control

schaefi:
code is great :+1:

doc/man/azurectl::compute::vm.md
rjschwei: Agreed, this is a good place to provide information about the waagent setting.

bear454: How about:

## __--custom-data=string-or-file__

A string of data, or a path to a file whose contents will be injected into the new virtual machine.

The Azure Service API will base64-encode the custom data during transport. During provisioning, waagent will store this data on the VM as both an attribute of __/var/lib/waagent/ovf-env.xml__ and as the sole contents of __/var/lib/waagent/CustomData__. By default, waagent will store the data on the VM as it was encoded during transport; the contents will need to be base64-decoded in order to access the original custom data. Waagent can be configured to decode the custom data and write out the original data, by changing the __Provisioning.DecodeCustomData__ attribute in __/etc/waagent.conf__.

Note: customdata is limited to 64K; using a file larger than 64K will fail.

rjschwei: The only thing I would change is the beginning. In the end the user doesn't care where or by what the data is encoded thus I'd say.

"""
The provided data or file content will be base64-encoded. During provisioning ......
"""

bear454: Updated in 2bedc1a.

PR #184, opened 2016-08-22
Data disk create

schaefi:
Refactor compute data-disk create command

* allow creation of data disks without a running instance in the background

Example use

```
$ azurectl --account publish compute data-disk create --identifier ms-bob

Created disk: ms-bob-data-disk-2016-08-22T09_15_25.950289

$ azurectl --account publish compute data-disk show \
    --disk-name ms-bob-data-disk-2016-08-22T09_15_25.950289

$ azurectl --account publish compute data-disk attach \
    --cloud-service-name suse-products --instance-name suma \
    --disk-name ms-bob-data-disk-2016-08-22T09_15_25.950289

$ azurectl --account publish compute data-disk detach \
    --cloud-service-name suse-products --instance-name suma --lun 0

$ azurectl --account publish compute data-disk delete \
    --disk-name ms-bob-data-disk-2016-08-22T09_15_25.950289
```

Operations are in async mode, there might be some wait state after detach before you can delete.
Passing on --wait to detach therefore could block for quite some time

### DIFF
--- a/azurectl/commands/compute_data-disk.py
+++ b/azurectl/commands/compute_data-disk.py
@@ -16,7 +16,7 @@ Data-disks are virtual disks attached to a virtual machine, backed by a virtual
 hard disk (VHD) image in Azure storage.
 
 usage: azurectl compute data-disk -h | --help
-       azurectl compute data-disk create --identifier=<name>
+       azurectl compute data-disk create --disk-basename=<name>
            [--size=<disk-size-in-GB>]
            [--label=<label>]
        azurectl compute data-disk delete --disk-name=<name>
@@ -40,7 +40,7 @@ commands:
     create
         create a new, empty data disk. The data disk vhd file will be
         created using the following naming schema:
-        <identifier>-data-disk-<utctime>. The default data disk size
+        <disk-basename>-data-disk-<utctime>. The default data disk size
         is set to 10GB
     delete
         delete the specified data disk. The call will fail if the disk
@@ -62,8 +62,8 @@ options:
         name of the cloud service where the virtual machine may be found
     --disk-name=<name>
         name of the data disk as registered in the image repository
-    --identifier=<name>
-        identifier string used as part of the complete data disk name.
+    --disk-basename=<name>
+        data disk basename used as part of the complete data disk name.
         Usually this is set to the instance name this data disk should be
         attached to later
     --instance-name=<name>
@@ -146,7 +146,7 @@ class ComputeDataDiskTask(CliTask):
 
     def __create(self):
         self.data_disk.create(
-            self.command_args['--identifier'],
+            self.command_args['--disk-basename'],
             self.command_args['--size'],
             self.command_args['--label']
         )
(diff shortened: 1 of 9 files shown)


azurectl/commands/compute_data-disk.py
```diff
@@ -16,13 +16,9 @@
 hard disk (VHD) image in Azure storage.
 
 usage: azurectl compute data-disk -h | --help
-       azurectl compute data-disk create --cloud-service-name=<name> --size=<disk-size-in-GB>
-           [--instance-name=<name>]
+       azurectl compute data-disk create --identifier=<name>
```
bear454: Why add a new arg (`--identifier`) instead of reusing `--disk-name` ?

bear454:
Aside from the question about args, the code looks great. Thanks!

azurectl/commands/compute_data-disk.py
schaefi: because the disk-name is build out of a base name (I call identifier) followed by kind of a type specification we call 'data-disk' followed by a time stamp which is the utc time information. In the other invocation types where the user has to provide the full name of the data-disk we use the --data-disk option but for create we only provide a part of the full name

schaefi: I also tried to explain  that with the initial comment

schaefi:
I'm open to change the argument if we find a better / more expressive name

azurectl/commands/compute_data-disk.py
rjschwei: Lets call it disk-basename as the provided information becomes part of the generated name. It is not an identifier in the sense that it would identify something that already exists or identifies something that is create because it is only a part of the name.

azurectl/commands/compute_data-disk.py
```diff
@@ -42,9 +38,10 @@
 
 commands:
     create
-        create a new, empty data disk attached to the specified instance.
-        The data disk vhd file will be created using the following naming
-        schema: <instance-name|cloud-service-name>-data-disk-<utctime>
+        create a new, empty data disk. The data disk vhd file will be
+        created using the following naming schema:
+        <identifier>-data-disk-<utctime>. The default data disk size
```
rjschwei: <basename>-data-disk-<utctime>

I also think we should give an example as "utctime" could mean a lot of things and the format can be widely different, i.e. it could be YYYYMMDDHHMMSS or HHMMSS, and example will clarify all of those questions.

The other question would be, why we cannot simply use a name the user gives us and why we have to create this expanded name?

schaefi: I like --disk-basename, but let's first clarify your second question

> why we cannot simply use a name the user gives us

I think this a result of the former implementation where the add_data_disk() api call was used which required to pass along an instance reference. The Azure service created the disk for us whereas they use the provided instance name as disk base name and created a disk-name out of it which they directly attach to the instance. We try to stay compatible with the naming here when we create a data disk which is not directly attached to an instance

I actually like that we do that but of course it's not required

PR #185, opened 2016-09-07
SAS URLs for disk images

bear454:


doc/man/azurectl::storage::disk.md
```diff
@@ -29,6 +34,13 @@ command is mainly for uploading XZ-compressed VHD (Virtual Hard Drive) disk
 images in order to register an Azure operating system image from it at a later
 point in time.
 
+## __sas__
+
+Generate a Shared Access Signature (SAS) URL allowing limited access to
+a disk image, without requiring an access key.
+See https://azure.microsoft.com/en-us/documentation/articles/storage-dotnet-shared-access-signature-part-1/
+for more information on shared access signatures.
+
```
schaefi: Can you write this in one line, the markup into the manual page requires this to auto-expand the texts when viewing it as a man page. We should switch to sphinx and ReST docs in the future that's way more convenient. But for now it would be nice if you can avoid the CR's. Thanks

schaefi:
looks great :+1: just the minor issue with the man page docs I mentioned

doc/man/azurectl::storage::disk.md
bear454: We should switch to `groff` and not bother with all the extra dependencies :stuck_out_tongue: 

Fixed in 11ff611.

PR #186, opened 2016-09-14
Reserved ip association

schaefi:
Add support for IP to cloud service associations

```
azurectl compute reserved-ip associate --name my_ip --cloud-service-name my_service
azurectl compute reserved-ip disassociate --name my_ip --cloud-service-name my_service
```

### DIFF
--- a/azurectl/commands/compute_reserved-ip.py
+++ b/azurectl/commands/compute_reserved-ip.py
@@ -18,13 +18,13 @@ be made. Reservations are identified by a user-defined name.
 usage: azurectl compute reserved-ip -h | --help
        azurectl compute reserved-ip create --name=<reserved-ip-name>
            [--wait]
-       azurectl compute reserved-ip list
-       azurectl compute reserved-ip show --name=<reserved-ip-name>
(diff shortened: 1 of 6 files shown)


azurectl/commands/compute_reserved-ip.py
```diff
@@ -22,9 +22,17 @@
...
+           [--wait]
+       azurectl compute reserved-ip disassociate --name=<reserved-ip-name> --cloud-service-name=<name>
+           [--wait]
        azurectl compute reserved-ip help
 
 commands:
+    associate
+        associate an existing reserved IP address to a deployment
+    disassociate
+        disassociate an existing reserved IP address from the given deployment
```
rjschwei: I think we agreed that options would be listed according to CRUD, IMHO associate and disassociate would fall into the U (updated category) and thus should be listed after "create".

If we agreed on alpha order then disassociate should be after create

rjschwei:
Other than the order of the options this LGTM

azurectl/commands/compute_reserved-ip.py
schaefi: I've changed the order, thanks

PR #190, opened 2016-09-20
Document: storage disk upload skips zero'd blocks

colemickens:
I started going down a different path for the blob upload until I skimmed the source and saw that you do in fact skip empty blocks. Nice feature, just thought it'd be good to document it in the man page.


schaefi:
Thanks :+1:

PR #191, opened 2016-09-21
add --show-in-gui and --recommended-vm-size to compute image update

colemickens:
This adds the ability to set `show_in_gui` and `recommended_vm_size` for compute images.

This won't work 100% yet because of these bugs:

* https://github.com/Azure/azure-sdk-for-python/issues/793 - recommended_vm_size is not populated
* https://github.com/Azure/azure-sdk-for-python/issues/794 - `show_in_gui` updates do not take effect

But, I can use `--show-in-gui` with a locally patched azure-storage-python package, and the recommended_vm_Size update works... but it will get clobbered on any subsequent updates because of how you've implemented the update logic.

So there's still value in merging this before the python-sdk fixes are implemented/merged.

### DIFF
--- a/azurectl/instance/image.py
+++ b/azurectl/instance/image.py
@@ -268,6 +268,8 @@ class Image(object):
             if value is not None:
                 if '_date' in name:
                     value = self.__convert_date_to_azure_format(value)
+                if 'show_in_gui' in name:
+                    value = value.lower() in ("yes", "true", "t", "1")
                 Defaults.set_attribute(image, name, value)
         return image
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.1.1
+current_version = 2.2.0
 commit = True
 tag = True
 
--- a/azurectl/commands/compute_image.py
+++ b/azurectl/commands/compute_image.py
@@ -35,6 +35,8 @@ usage: azurectl compute image -h | --help
            [--privacy-uri=<privacy_uri>]
            [--published-date=<date>]
            [--small-icon-uri=<small_icon_uri>]
+           [--show-in-gui=<show_in_gui>]
+           [--recommended-vm-size=<recommended_vm_size>]
        azurectl compute image publish --name=<imagename>
            [--private]
            [--msdn]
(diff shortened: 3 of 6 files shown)


schaefi:
I'm ok with adding the additional image metadata fields even though the SDK has issues with it, it will self heal once the SDK is fixed

However, your patch is missing the required unit test coverage. Could you add the tests too ? Thanks

azurectl/instance/image.py
```diff
@@ -267,7 +267,9 @@ def __decorate_os_image_attributes_for_update(self, image, update_record):
         for name, value in ordered_update_record.iteritems():
             if value is not None:
                 if '_date' in name:
-                    value = self.__convert_date_to_azure_format(value)
+                   value = self.__convert_date_to_azure_format(value)
+                if 'show_in_gui' in name:
+		   value = value.lower() in ("yes", "true", "t", "1")
```
schaefi: we have spacing issues here, 4 spaces should be used for indentation. flake will also complain about this

colemickens: Thanks, couldn't figure out why it was showing me that line in a diff, but I was viewing side-by-side. Will fix!

lmazuel:
FYI @schaefi azure-servicemanagement-legacy 0.20.5 released with the bug fixes related to this one.
https://pypi.python.org/pypi/azure-servicemanagement-legacy/0.20.5

colemickens:
@schaefi Tests added. Not sure if I handled the boolean flag correctly.

Also, not sure if I need to make a backward-compat change for the `show_in_gui` flag. It's unclear to me if it's going to default to `False` now, whereas it defaulted to `True` when being omitted, apparently.

schaefi:
There seems to be still a flake8 issue

From the travis log above

```
flake8
./azurectl/instance/image.py:270:20: E111 indentation is not a multiple of four
```

You can prevent travis from complaining if you run the flake8 and unit test targets in your development environment pior to creating the pull request. Also see:

* https://github.com/SUSE/azurectl#contributing
* https://github.com/SUSE/azurectl#testing

Hope that helps

schaefi:
There seems to be still a flake8 issue

From the travis log above

```
flake8
./azurectl/instance/image.py:270:20: E111 indentation is not a multiple of four
```

You can prevent travis from complaining if you run the flake8 and unit test targets in your development environment pior to creating the pull request. Also see:

* https://github.com/SUSE/azurectl#contributing
* https://github.com/SUSE/azurectl#testing

Hope that helps

colemickens:
Hi @schaefi. Thanks for the pointers and patience.

I've updated the PR and added a new commit that bumps the version. The tests and flake8 are passing, please take a look.

schaefi:
great work, thanks :+1:

colemickens:
Thanks for reviewing and merging! I'm assuming pushing the new version tag is something you do manually? Can you possibly do that as well? Thanks!

schaefi:
>    Thanks for reviewing and merging! I'm assuming pushing the new version
>    tag is something you do manually? Can you possibly do that as well?
>    Thanks!

sure, you will find roling releases here:

    http://download.opensuse.org/repositories/Cloud:/Tools

I will make a submission today, after I've finished my current Azure
upload and testing.

PR #195, opened 2016-10-31
Fixup data disk creation

schaefi:
The data disk file must be of the specified size in sparse
containing the vhd metadata at the end of the disk. Azure
only supports fixed size VHD's


schaefi:
Tested the fix in an instance still running here:

https://manage.windowsazure.com/@azurerdsuse.onmicrosoft.com#Workspaces/VirtualMachineExtension/VirtualMachine/ce73a2b0-d2e7-4ff6-b987-b32d6908de4e,ms-bootpart-test,ms-bootpart-test,ms-bootpart-test/dashboard

I haven't observed a shutdown of the instance

PR #196, opened 2016-12-13
Only upload vhd metadata on data-disk create

bear454:
The only content in an empty Azure data-disk is the fixed VHD footer, so instead of creating and reading
an arbitary-sized empty file, just upload the footer data.

### DIFF
--- a/azurectl/instance/data_disk.py
+++ b/azurectl/instance/data_disk.py
@@ -14,7 +14,6 @@
 from azure.common import AzureMissingResourceHttpError
 from azure.storage.blob.pageblobservice import PageBlobService
 from datetime import datetime
-from tempfile import NamedTemporaryFile
 from builtins import bytes
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


schaefi:
Are you sure about this ? meaning have you tested it and the VM did not run into the issue I have fixed with commit #8227893cc12cb1b5b13717ec56eb32f60a5b4cb6

In addition does the attached disk really have the requested size available for writing on the VM ?

I'm very doubtful about vhdfixed disk formats and I know the current implementation works fine because it has been tested for real in all combinations I could imagine, thus let's try hard to not break it.

I trust your coding, forgive me when I'll ask stupid questions

bear454:
Yes, I tested many times, and with the final image brought up two new servers each with two disks. 

Look at the variance in code... we weren't constructing a VHD disk image... just truncate'ing to a specific length, then writing the footer. Performance issues came from the upload portion, which was reading through all the zero blocks (often in our case nearly a TB of them). You can see, I didn't change the footer we were writing to disk, and if you follow the process through page_blob, you'll see in the end it did before only one block upload, and does now only one block upload.

schaefi:
yes makes all sense, sorry for being concerned

PR #202, opened 2017-01-18
Return dictionary status from request_status.

smarlowucf:
The [get_operation_status](https://github.com/Azure/azure-sdk-for-python/blob/f8dc98625404b75e5d6365ad30c04abc90ee751a/azure-servicemanagement-legacy/azure/servicemanagement/servicemanagementclient.py#L320) command from azure SDK is returning an [Operation](https://github.com/Azure/azure-sdk-for-python/blob/f8dc98625404b75e5d6365ad30c04abc90ee751a/azure-servicemanagement-legacy/azure/servicemanagement/models.py#L747) object not a string.

Return a dictionary with the result status plus the error message if the request has failed.
- I don't think the SDK Operation object should be passed through CliTask.request_status()
- But I do think passing the error message on failure is useful

Hence why I chose to return a dictionary rather than just the status string.

### DIFF
--- a/azurectl/commands/base.py
+++ b/azurectl/commands/base.py
@@ -101,4 +101,10 @@ class CliTask(object):
         if self.account:
             service = self.account.get_management_service()
             request_result = RequestResult(request_id)
-            return format(request_result.status(service))
+            result_status = request_result.status(service)
+            status = {'result': result_status.status}
+
+            if result_status.status == 'Failed':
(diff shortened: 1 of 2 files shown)


schaefi:
The reason why I choose to return the object instance was that you will get an information where in the SDK code the exception was raised. I agree this was more useful from a developer perspective than from a user perspective. Ok so let's change this, thanks for the patch

PR #203, opened 2017-01-30
Added compute vm shutdown command

### LINKED ISSUE #200: azurectl should be able to stop an instance
At present the vm subcommand only allows to delete and create VMs, but one should be able to stop an instance as well.

schaefi:
Allows to shut down a virtual machine instance. Also taking
into account if the used resources should be deallocated or
not. This Fixes #200

### DIFF
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -75,7 +75,8 @@ options:
         in a shutdown request, shuts down the Virtual Machine and releases
         the compute resources. You are not billed for the compute resources
         that this Virtual Machine uses. If a static Virtual Network IP
-        address is assigned to the Virtual Machine, it is reserved
+        address is assigned to the Virtual Machine the status of the IP
+        address is changed to become a reserved IP address
     --fingerprint=<thumbprint>
         thumbprint of an already existing certificate in the
         cloud service used for ssh public key authentication
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -32,6 +32,10 @@ usage: azurectl compute vm -h | --help
            [--wait]
        azurectl compute vm regions
        azurectl compute vm show --cloud-service-name=<name>
+       azurectl compute vm shutdown --cloud-service-name=<name>
+           [--instance-name=<name>]
+           [--deallocate-resources]
+           [--wait]
        azurectl compute vm types
        azurectl compute vm delete --cloud-service-name=<name>
            [--instance-name=<name>]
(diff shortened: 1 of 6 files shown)


schaefi:
I've associated role_name as to be the instance_name with regards to the sdk implementation. The shutdown sequence needs further real live testing before we merge

azurectl/commands/compute_vm.py
```diff
@@ -65,6 +71,11 @@
     --custom-data=<string-or-file>
         a string of data or path to a file that will be injected into the new
         virtual machine
+    --deallocate-resources
+        in a shutdown request, shuts down the Virtual Machine and releases
+        the compute resources. You are not billed for the compute resources
+        that this Virtual Machine uses. If a static Virtual Network IP
+        address is assigned to the Virtual Machine, it is reserved
```
rjschwei: You probably mean "...is preserved" mening it is not givenback to the framework.

doc/man/azurectl::compute::vm.md
```diff
@@ -79,6 +88,10 @@ The provided data or file content will be base64-encoded. During provisioning, w
 
 Note: customdata is limited to 64K; using a file larger than 64K will fail.
 
+## __--deallocate-resources__
+
+In a shutdown request, shuts down the Virtual Machine and releases the compute resources. You are not billed for the compute resources that this Virtual Machine uses. If a static Virtual Network IP address is assigned to the Virtual Machine, it is reserved.
+
```
rjschwei: See comment for help.

azurectl/commands/compute_vm.py
schaefi: Ahm no I have taken this information from the SDK code. I think this means if you have used a static ip and drop the resource on shutdown it is turned into what Azure calls an "reserved ip" So at reboot the instance gets the same ip back

doc/man/azurectl::compute::vm.md
schaefi: see answer above 😄 I'm open to change the wording if you think it's not clear. But I think for users of azure who have an understanding of "reserved ip's" it might be ok. Thoughts ?

schaefi: we also have the command 'azurectl compute reserved-ip'

azurectl/commands/compute_vm.py
rjschwei: OK, then lets clarify that in our help message. Just because they write stuff that is confusing does not mean we have to follow suit.

rjschwei: Suggestion:

"...address is assigned to the Virtual Machine the status of the IP address is changed to become a reserved IP address."

PR #207, opened 2017-02-16
Fix methods if no endpoints exist

smarlowucf:
Currently the instance endpoint methods all fail if the cloud service and/or instance have no endpoints. Instead of returning an empty list Azure is returning a None type for input_endpoints value.

- Add checks for None
- Add check for endpoint in delete and only delete if it exists
- Create new endpoints list on creation of an endpoint if value is None

### DIFF
--- a/azurectl/instance/endpoint.py
+++ b/azurectl/instance/endpoint.py
@@ -46,7 +46,11 @@ class Endpoint(object):
             raise AzureEndpointListError(
                 '%s: %s' % (type(e).__name__, format(e))
             )
-        results = config.input_endpoints if config.input_endpoints else []
+
+        if config.input_endpoints:
+            results = config.input_endpoints
+        else:
+            results = []
         return [self.__decorate(result) for result in results]
 
     def show(self, name):
@@ -57,7 +61,8 @@ class Endpoint(object):
                 '%s: %s' % (type(e).__name__, format(e))
             )
 
-        # If there are no endpoints the list is none not []
+        # If there are no endpoints the input_endpoints
+        # attribute is None.
         if config.input_endpoints:
             for endpoint in config.input_endpoints:
                 if endpoint.name == name:
@@ -88,8 +93,8 @@ class Endpoint(object):
                 enable_direct_server_return=False,
             )
 
-            # If there are no endpoints the list is none so create
-            # new list of endpoints.
+            # If there are no endpoints the input_endpoints
+            # attribute is None. Thus create and set new instance.
             if not config.input_endpoints:
                 config.input_endpoints = ConfigurationSetInputEndpoints()
 
--- a/azurectl/instance/endpoint.py
+++ b/azurectl/instance/endpoint.py
@@ -121,19 +121,19 @@ class Endpoint(object):
             config = self.__get_network_config_for_role(role)
 
             if not config.input_endpoints:
-                raise Exception(
+                raise AzureEndpointDeleteError(
                     "No endpoints found."
                 )
 
-            for i, endpoint in \
+            for index, endpoint in \
                     enumerate(config.input_endpoints.input_endpoints):
                 if endpoint.name == name:
-                    del config.input_endpoints.input_endpoints[i]
+                    del config.input_endpoints.input_endpoints[index]
                     break
             else:
                 # If for loop finishes normally no endpoint
                 # exists that matches the name.
-                raise Exception(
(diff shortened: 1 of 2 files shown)


azurectl/instance/endpoint.py
```diff
@@ -43,7 +46,7 @@ def list(self):
             raise AzureEndpointListError(
                 '%s: %s' % (type(e).__name__, format(e))
             )
-        results = config.input_endpoints
+        results = config.input_endpoints if config.input_endpoints else []
         return [self.__decorate(result) for result in results]
```
rjschwei: I hate it that the Python implementers decided that it was a good idea to copy bad syntax constructs from Perl. Postfix if sucks! Please break this into

if config.input_endpoints:
    results = config.input_endpoints
else:
    results = []

Perl best practices page 93 "Use block if, not postfix if" it was a bad idea in Perl and it remains a bad idea even if some people think it is great to copy bad coding style idea from other languages.

azurectl/instance/endpoint.py
```diff
@@ -53,17 +56,17 @@ def show(self, name):
...
-                result = endpoint
-                break
-        if result:
-            return self.__decorate(result)
-        else:
-            raise AzureEndpointShowError(
-                "No endpoint named '%s' was found." % name
-            )
+
+        # If there are no endpoints the list is none not []
```
rjschwei: This comment is incorrect, None is a data type thus "the list is none not []" is misleading

# If there are no endpoints the return is None rather than an empty list

is an option as a comment

azurectl/instance/endpoint.py
```diff
@@ -84,6 +87,12 @@ def create(
                 idle_timeout_in_minutes=idle_timeout,
                 enable_direct_server_return=False,
             )
+
+            # If there are no endpoints the list is none so create
+            # new list of endpoints.
```
rjschwei: Ditto on this comment, a list cannot be None

azurectl/instance/endpoint.py
smarlowucf: It's helpful in code golf though :wink:

azurectl/instance/endpoint.py
smarlowucf: Yes, bad wording. Will clean up both comments.

azurectl/instance/endpoint.py
```diff
@@ -105,11 +119,23 @@ def delete(self, name):
...
             role = self.__get_role()
             config = self.__get_network_config_for_role(role)
-            new_endpoints = []
-            for endpoint in config.input_endpoints.input_endpoints:
-                if endpoint.name != name:
-                    new_endpoints.append(endpoint)
-            config.input_endpoints.input_endpoints = new_endpoints
+
+            if not config.input_endpoints:
+                raise Exception(
```
schaefi: can we raise an AzureEndpointDeleteError please

azurectl/instance/endpoint.py
```diff
@@ -105,11 +119,23 @@ def delete(self, name):
...
+                    "No endpoints found."
+                )
+
+            for i, endpoint in \
```
schaefi: Can we make this more explicit, 'i' vs. 'index' or any other variable name which is more expressive

azurectl/instance/endpoint.py
```diff
@@ -105,11 +119,23 @@ def delete(self, name):
...
+                    enumerate(config.input_endpoints.input_endpoints):
+                if endpoint.name == name:
+                    del config.input_endpoints.input_endpoints[i]
+                    break
+            else:
+                # If for loop finishes normally no endpoint
+                # exists that matches the name.
+                raise Exception(
```
schaefi: can we raise an AzureEndpointDeleteError please

azurectl/instance/endpoint.py
smarlowucf: :+1:

azurectl/instance/endpoint.py
smarlowucf: Yeah, the reason I went with Exception is we re-raise with AzureEndpointDeleteError. So the message looks like:

`ERROR: AzureEndpointDeleteError: 'Exception: "No endpoints found."'`

instead of

`ERROR: AzureEndpointDeleteError: 'AzureEndpointDeleteError: "No endpoints found."'`

But I could add an except block to re-raise the exception as is...

PR #208, opened 2017-02-16
Use `compute data-disk attach` for either data-disks or VHD page blobs.

### LINKED ISSUE #206: attaching data-disks requires different API interactions
I found while moving data-disks from one machine to another, that our current implementation cannot attach an existing, populated vhd file (page blob) to an instance.

The API is a bit ambiguous about it, with two arguments for supplying the URL of the vhd blob; we need to determine which arguments work in which cases, and code an appropriate fix.

The next comment will include a test matrix; I'll run through each scenario using the `media_link` argument, then with the `source_media_link` argument, and report results.

bear454:
Resolves #206 .


azurectl/instance/data_disk.py
```diff
@@ -153,10 +152,38 @@ def attach(
...
+            if disk_name:
+                args = {
+                    'disk_name': disk_name
+                }
+            else:
+                args = {
+                    'disk_name': self.__disk_name_without_extension(blob_name),
+                    'source_media_link': self.__data_disk_url(blob_name)
+                }
+        else:
```
rjschwei: This condition is superfluous as the arguments were validated and we should never reach this code. Please remove

azurectl/instance/data_disk.py
```diff
@@ -153,10 +152,38 @@ def attach(
...
+            }
+        elif not disk_name and blob_name:
+            # find data-disk name for blob_name,
+            # or create a new data-disk for blob_name
+            disk_name = self.__disk_name_for_blob_name(
+                blob_name,
+                self.service.list_disks()
+            )
+            if disk_name:
+                args = {
```
rjschwei: __disk_name_for_blob_name should always return a valid string, thus the if disk_name: test is superfluous

azurectl/instance/data_disk.py
```diff
@@ -153,10 +152,38 @@ def attach(
...
+            else:
+                args = {
+                    'disk_name': self.__disk_name_without_extension(blob_name),
+                    'source_media_link': self.__data_disk_url(blob_name)
+                }
+        else:
+            raise AzureDataDiskAttachError(
+                "Neither disk_name nor blob_name was supplied"
+            )
+
```
rjschwei: Should there be a test here to see if the disk-name exists and the blob-name has been specified? Or can we depend on the API to complain if we try to regenerate data that all ready exists.

So if I have --disk-name foo --blob-name my-blob.vhd but data disk with name foo already exists, will the API generate an error when we try to create A or will it clobber what exists?

azurectl/instance/data_disk.py
```diff
@@ -228,6 +255,23 @@ def __data_disk_url(self, filename):
...
+        for extension in extensions:
+            if name.endswith(extension):
+                return name[:-len(extension)]
+
+    def __disk_name_for_blob_name(self, blob_name, disks):
+        disk_url = self.__data_disk_url(blob_name)
+        try:
+            disk_name, = (
+                disk.name for disk in disks if disk.media_link == disk_url
```
rjschwei: postfix for and if, just as ugly and bad as in other language

azurectl/instance/data_disk.py
```diff
@@ -228,6 +255,23 @@ def __data_disk_url(self, filename):
             filename
         )
 
+    def __disk_name_without_extension(self, name):
+        extensions = ['.vhd', '.vhdfixed']
+
```
rjschwei: Is there some intrinsic restriction that guarantees that only .vhd and .vhdfixed would be used as extensions?

Is there no circumstance where I can end up with .foo?
If I can end up with ".foo" then we should use split('.')[0] and use simply the first part of a name that is separated by '.'

rjschwei:
General comment since --disk-name moves from required to optional we make an API change and push the version to 3.0.0. I am OK with that, just wanted to point that out.

azurectl/commands/compute_data-disk.py
```diff
@@ -20,8 +20,10 @@
            [--size=<disk-size-in-GB>]
            [--label=<label>]
        azurectl compute data-disk delete --disk-name=<name>
-       azurectl compute data-disk attach --cloud-service-name=<name> --disk-name=<name>
+       azurectl compute data-disk attach --cloud-service-name=<name>
            [--instance-name=<name>]
+           [--disk-name=<name>]
+           [--blob-name=<name>]
            [--label=<label>]
```
schaefi: This can be solved the docopt way

```python
azurectl compute data-disk attach --cloud-service-name=<name> (--disk-name=<name>|--blob-name=<name>)
```

azurectl/commands/compute_data-disk.py
```diff
@@ -190,11 +195,14 @@ def __delete(self):
         log.info('Deleted data disk %s', self.command_args['--disk-name'])
 
     def __attach(self):
+        self.validate_at_least_one_argument_is_set('--disk-name', '--blob-name')
```
schaefi: This code would not be required if we let docopt check it

azurectl/instance/data_disk.py
bear454: It shouldn't; I'll be more explicit about the function.

azurectl/instance/data_disk.py
bear454: The arguments accessible outside the CLI, by importing azurectl, or through the `compute shell` command.

I welcome recommendations, though, on a better way to handle either/or/both arguments.

azurectl/instance/data_disk.py
bear454: No there's no intrinsic restriction, but looking for the period separator isn't a great choice either: periods are allowed in the names, and I know some users (us) use them in version numbers, for example, and since Azure doesn't apply the vhd extension to root disk images, it could be messy.

This was more about maintaining consistency with azurectl (`data-disk create` appends ".vhd" to the disk name - see https://github.com/SUSE/azurectl/blob/master/azurectl/instance/data_disk.py#L215 ).

I'd gladly welcome requests too add more extensions, but I think splitting on the dot is too dangerous.

azurectl/commands/compute_data-disk.py
bear454: The arguments aren't mutually exclusive: either _or_ both is acceptable.

It would have to be `(--disk-name=<name>|--blob-name=<name>|--disk-name=<name> --blob-name=<name>)` which is just ugly and I'm not sure even works.

azurectl/instance/data_disk.py
bear454: Resolved in bc48685 .

azurectl/commands/compute_data-disk.py
bear454: I don't see `(['--disk-name', '--blob-name'])` as more intuitive - just clunkier, but I'll take this one on _style_ (and keep @rjschwei happy ;-)

Resolved in cd21a1a.

azurectl/instance/data_disk.py
bear454: Good catch; testing...

bear454: The API complains:

```
ERROR: AzureRequestError: u'Operation a5a090dd89587c24adef5410c1592daf failed. The VHD https://suserdwestus2.blob.core.windows.net/vm-images/jmason-test1-attach.vhd is already registered with image repository as the resource with ID jmason-test1-attach. (ConflictError)'
```

azurectl/instance/data_disk.py
bear454: You just don't like one-liners :stuck_out_tongue_closed_eyes: 

Expanded in 4eb061a.

azurectl/instance/data_disk.py
smarlowucf: Would a [right split](https://docs.python.org/2/library/stdtypes.html#str.rsplit) with a maxsplit of 1 work here?

`rsplit('.', 1)[0]`

That would start split from right and only split the first period.

bear454: Here's an example: the root VHD disk - a page blob in storage - generated by Azure for an SMT server, which follows the pattern `#{cloud-service-name}_instance_#{instance-name}_image_#{image-name}`:

> azure-smt1-germany-central_instance_azure-smt12-1-germany-central_image_germany-central-sles12-sp1-azure-smtserver.x86-64-0.1.18-build1.3

If we wanted to mount this on another instance, for example, for forensics, it wouldn't make any sense to drop the ".3" off the end.

In contrast, if you add a data-disk through the Azure portal, it _does_ get the ".vhd" extension, but the name of the VHD file is _ignored_ in setting the data-disk-name, which is derived from the instance name, LUN, and a timestamp. :/

azurectl/commands/compute_data-disk.py
smarlowucf: Not sure if this is a better approach but it could be achieved in docopt with two usage statements.

```
azurectl compute data-disk attach --cloud-service-name=<name>
           [--instance-name=<name>]
           --disk-name=<name>
           [--blob-name=<name>]
           [--label=<label>]
           [--lun=<lun>]
           [--no-cache|--read-only-cache|--read-write-cache]
           [--wait]
  azurectl compute data-disk attach --cloud-service-name=<name>
           [--instance-name=<name>]
           [--disk-name=<name>]
           --blob-name=<name>
           [--label=<label>]
           [--lun=<lun>]
           [--no-cache|--read-only-cache|--read-write-cache]
           [--wait]
```

bear454: I've confirmed... this is the "docopt way" of dealing with this:

```
       azurectl compute data-disk attach --cloud-service-name=<name>
           (--disk-name=<name> | --blob-name=<name> | --disk-name=<name> --blob-name=<name>)
           [--instance-name=<name>]
           [--label=<label>]
           [--lun=<lun>]
           [--no-cache|--read-only-cache|--read-write-cache]
           [--wait]
```

I'm fine with going this way; the benefit is dropping the argument validator, but I want some consensus on that decision. @rjschwei ? @schaefi ? @smarlowucf ?

smarlowucf: :+1: for the single line approach `(--disk-name=<name> | --blob-name=<name> | --disk-name=<name> --blob-name=<name>)`

azurectl/instance/data_disk.py
rjschwei: OK, so the method name should change. "..._without_extension" is generic, but what we appear to really want is to strip stuff that gets added by the platform, or our special extension of whdfixed. So lets name the method accordingly.

_strip_platform_extension

may be an option.

azurectl/commands/compute_data-disk.py
rjschwei: Not fond of the repetition I think it is confusing. Consider that none of us, even @schaefi who works the most with azurectl look at this every day. Neither do we work with docopt everyday. So in 6 months do we have to lookup the given syntax in the documentation?

The validator code makes the intend clear, I am not so certain the docopt implementation will in 6 months.

However, I am fine with using the docopt implementation, in the end I have contributed little actual code to azurectl.

azurectl/instance/data_disk.py
rjschwei: Now I get it, sorry, inline loops and postfix conditions, as I keep saying Perl Best Practices page 93 and following

azurectl/commands/compute_data-disk.py
schaefi: I agree to what Robert said, the repetition is confusing. Let's go with the original code and keep the argument validator

azurectl/instance/data_disk.py
bear454: Renamed in 935032a .

bear454:
@schaefi can you do one more overall review?

azurectl/instance/data_disk.py
```diff
@@ -228,6 +255,22 @@ def __data_disk_url(self, filename):
...
+    def __strip_platform_extension(self, name):
+        extensions = ['.vhd', '.vhdfixed']
+
+        for extension in extensions:
+            if name.endswith(extension):
+                return name[:-len(extension)]
+
+    def __find_existing_disk_name_for_blob_name(self, blob_name, disks):
+        '''if a data-disk object exists for the given blob name, find it;
+        if not, return None as a signal to generate one.'''
```
schaefi: nitpick but we use the docstring format

```
"""
    comment
"""
```

would be consistent with the rest of the code

azurectl/utils/validations.py
```diff
@@ -57,3 +57,14 @@ def validate_date(self, field_name, value):
...
+    @classmethod
+    def validate_at_least_one_argument_is_set(self, command_args, keys=[]):
+        values = [command_args[key] for key in keys]
+        if any(values):
+            return True
+        else:
+            raise AzureInvalidCommand(
+                'One of the following arguments must be supplied: %s' %
+                ', '.join(keys)
+            )
```
schaefi: The python guru's reviewing my code always complains about this and tells me to use format

```python
'One of the following arguments must be supplied: {args}'.format(args=', '.join(keys))
```

schaefi: Feel free to decide to make use of it, I'm fine with both

schaefi:
@bear454 I looked over the code and apart from the two minor suggestions I'm fine with it, looks great to me. It would be nice if the commit list can be squashed before merge and there is also Robert's open review. It looks like he still has some concerns

azurectl/utils/validations.py
bear454: Since all of the validations use the same format, I'm going to leave it.

azurectl/instance/data_disk.py
bear454: Resolved in ee4f4f9.

bear454:
As requested, I'm going to close this PR, squash the fixes in, and resubmit.

PR #209, opened 2017-02-23
Use `compute data-disk attach` for either data-disks or VHD page blobs.

### LINKED ISSUE #206: attaching data-disks requires different API interactions

bear454:
Resolved #206

### DIFF
--- a/azurectl/commands/base.py
+++ b/azurectl/commands/base.py
@@ -91,6 +91,12 @@ class CliTask(object):
             self.command_args[cmd_arg]
         )
 
+    def validate_at_least_one_argument_is_set(self, keys=[]):
+        return Validations.validate_at_least_one_argument_is_set(
+            self.command_args,
+            keys
+        )
+
     def request_wait(self, request_id):
         if self.account:
             service = self.account.get_management_service()
--- a/azurectl/azurectl_exceptions.py
+++ b/azurectl/azurectl_exceptions.py
@@ -137,6 +137,10 @@ class AzureCustomDataTooLargeError(AzureError):
     pass
 
 
+class AzureDataDiskAttachError(AzureError):
+    pass
+
+
 class AzureDataDiskCreateError(AzureError):
     pass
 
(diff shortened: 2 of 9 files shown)


azurectl/commands/base.py
```diff
@@ -91,6 +91,12 @@ def validate_date(self, cmd_arg):
             self.command_args[cmd_arg]
         )
 
+    def validate_at_least_one_argument_is_set(self, keys=[]):
+        return Validations.validate_at_least_one_argument_is_set(
+            self.command_args,
+            keys
+        )
+
```
rjschwei: I am not certain this indirection adds value, i.e. at the call site of validate_at_least_one_argument_is_set one can also call the Validations.validate_at_least_one_argument_is_set without loss of clarity or intend. I think we can safe ourselves the extra hop.

rjschwei:
The extra indirection could also be considered style, thus I am not going to be adamant about making a code change.

bear454:
Not intentional indirection, but managing dependency injection, and scope of responsibility:

In this case, `Validation#validate_at_least_one_argument_is_set` requires as arguments a dictionary, and an array of keys, and ensures that at least one of `dictionary[key]` is assigned. This is easy to test :D

`CliTask#validate_at_least_one_argument_is_set` only requires as arguments an array of keys, as the dictionary is a class variable generated by `docopt`, and inherited by the specific command classes. So, the validation we perform in a command is _always_ against the CLI arguments, which makes sense and, as a bonus, simplifies the actual use of validation.

So... I'm going to leave it.

PR #210, opened 2017-03-03
Use absolute imports

bear454:
Relative imports fail when imported from a different scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,14 +378,6 @@ Run the check as follows:
 $ make flake8
 ```
 
-Running the application from source without the need to install it
-can be done as follows:
-
-```
-$ cd bin
-$ ./azurectl
-```
-
 The primary focus of testing is unit testing without verification of the
 integration. Therefore is is not required to have a Microsoft Azure account
 to run the tests or contribute to the project. When integration tests will be
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ the `test/unit` directory to see current implementations
 #### Class implementing desired functionality: mycmd.py
 
 ```python
-from azurectl_exceptions import (
+from azurectl.azurectl_exceptions import (
     exception_class_from_azurectl_exceptions
 )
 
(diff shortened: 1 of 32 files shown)


schaefi:
why do relative imports fail ? Actually that's not true. All azurectl code is placed below the 'azurectl' namespace, from there all relative imports are always correct. Thus using azurectl from other programs worked fine for me with either

```
import azurectl

or

from azurectl import ...
```

I don't see a reason why this was changed and merged without a clear error report or discussion
Not funny

bear454:
Well, while you may not have experienced it in the past, it is true; as part of the replaceSMT script, which is _my_first time importing azurectl into another python script, all relative imports failed to load.

There's no bug report here because it was discussed directly in chat with @rjschwei , who also reviewed & merged the code.

If you need additional support, be aware that according to https://wiki.python.org/moin/FutureProofPython  :

> this won't work, as all imports will be absolute by default
> **Don't use implicitly relative imports in new code.**

So if nothing else, consider it future proofing.

rjschwei:
On 03/06/2017 02:57 AM, Marcus Schäfer wrote:
> why do relative imports fail ? Actually that's not true. All azurectl
> code is placed below the 'azurectl' namespace, from there all relative
> imports are always correct. Thus using azurectl from other programs
> worked fine for me with either
> 
> |import azurectl or from azurectl import ... |
> 
> I don't see a reason why this was changed and merged without a clear
> error report or discussion
> Not funny
> 

Fair enough.

Unfortunately and for whatever reason the use of relative imports
slipped through all previous code reviews. There are known and well
documented issues with the Python import system. The import system is
complicated and getting consistent behavior is not necessarily easy,
consistent meaning across testing, "production" deployment and use as a
library.

For example the current implementation would fail if testing development
code outside a virtualenv if the path is not set properly to find the
the azurectl directory and thus all tests would fall over. On the other
hand the relative imports fail if azurectl is imported into some other
code, as @bear454 experienced. There is not really a good way to deal
with this. Using the absolute import at the very least provides us with
predictable behavior and allows azurectl code to be use in other scripts
that do not need the command line parsing functionality. Thus from that
perspective it increases the potential user base.

The relative imports should have never made it as far as they did. Thus
I opted for a quick merge to not block another project. Sorry for
stepping on your toes.

PR #211, opened 2017-03-10
Add update method to instance endpoint.

smarlowucf:
Add method and cli option to update an existing endpoint on an instance. All args can be changed except the endpoint name. This adds another arg --tcp to allow the choice of tcp, udp or None when updating. If neither option is chosen the current value is maintained.

### DIFF
--- a/azurectl/commands/compute_endpoint.py
+++ b/azurectl/commands/compute_endpoint.py
@@ -23,6 +23,13 @@ usage: azurectl compute endpoint -h | --help
            [--idle-timeout=<minutes>]
            [--udp]
            [--wait]
+       azurectl compute endpoint update --cloud-service-name=<name> --name=<name>
+           [--instance-name=<name>]
+           [--port=<port>]
+           [--instance-port=<port>]
+           [--idle-timeout=<minutes>]
+           [--udp | --tcp]
+           [--wait]
        azurectl compute endpoint list --cloud-service-name=<name>
            [--instance-name=<name>]
        azurectl compute endpoint show --cloud-service-name=<name> --name=<name>
@@ -35,6 +42,8 @@ usage: azurectl compute endpoint -h | --help
 commands:
     create
         add a new endpoint
+    update
+        update an existing endpoint
     delete
         remove an endpoint
     list
@@ -61,6 +70,8 @@ options:
         name of the endpoint, usually the name of the protocol that is carried
     --port=<port>
         port to open on the cloud service
+    --tcp
+        select TCP as the transport protocol for the endpoint. (update only)
     --udp
         select UDP as the transport protocol for the endpoint. If not specified,
         the default transport protocol is TCP
@@ -110,6 +121,8 @@ class ComputeEndpointTask(CliTask):
             self.__show()
         if self.command_args['create']:
             self.__create()
+        if self.command_args['update']:
+            self.__update()
         if self.command_args['delete']:
             self.__delete()
 
@@ -141,6 +154,31 @@ class ComputeEndpointTask(CliTask):
             ),
             ('udp' if self.command_args['--udp'] else 'tcp'),
             (self.command_args['--idle-timeout'] or '4')
+        )
+        if self.command_args['--wait']:
+            self.request_wait(request_id)
+        self.result.add(
+            'endpoint:' + self.command_args['--name'], request_id
+        )
+        self.out.display()
+
+    def __update(self):
+        if self.command_args['--udp']:
+            protocol = 'udp'
+        elif self.command_args['--tcp']:
+            protocol = 'tcp'
+        else:
+            protocol = None
+
+        request_id = self.endpoint.update(
+            self.command_args['--name'],
+            self.command_args['--port'],
+            (
+                self.command_args['--instance-port'] or
+                self.command_args['--port']
+            ),
+            protocol,
+            self.command_args['--idle-timeout']
         )
         if self.command_args['--wait']:
             self.request_wait(request_id)
--- a/azurectl/instance/endpoint.py
+++ b/azurectl/instance/endpoint.py
@@ -21,7 +21,8 @@ from azurectl.azurectl_exceptions import (
     AzureEndpointCreateError,
     AzureEndpointDeleteError,
     AzureEndpointListError,
-    AzureEndpointShowError
+    AzureEndpointShowError,
+    AzureEndpointUpdateError
 )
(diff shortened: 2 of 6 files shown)


azurectl/commands/compute_endpoint.py
```diff
@@ -23,6 +23,13 @@
            [--idle-timeout=<minutes>]
            [--udp]
            [--wait]
+       azurectl compute endpoint update --cloud-service-name=<name> --name=<name>
+           [--instance-name=<name>]
+           [--port=<port>]
+           [--instance-port=<port>]
+           [--idle-timeout=<minutes>]
+           [--udp | --tcp]
```
smarlowucf: Given azurectl is using the legacy service management api which [prevents creation of udp endpoints](https://github.com/Azure/azure-sdk-for-python/issues/334) I think it would be cleaner to remove the protocol option.

A udp endpoint cannot be created via azurectl or azure sdk. The only option would be to change an existing udp to tcp.

schaefi: It would be an incompatible change. I'm ok with deleting the --udp endpoint support but only if we don't need it. I think for our use case we only need to add tcp endpoints but I'm not sure. So let James and Robert add their thoughts.

From a technical perspective I think the only hurdle is that the request should not set an idle timeout for the setup of an udp endpoint. I didn't had a closer look, but we could also consider to add a patch in the sdk code to make this work

schaefi:
code looks great 👍

azurectl/commands/compute_endpoint.py
rjschwei: OK, I looked through the referenced thread and poked at the code a bit (azure sdk). It appears that the claim that endpoint creation for udp is being prevented is over stated. Based on the info in the issue it appears that endpoint creation for udp is possible just not straight forward. Now that leaves the question whether we handle it correctly or not, and if we do not handle it correctly whether we want to spend the effort to fix it or remove it.

So before we remove anything we first need to confirm that azurectl can indeed not create a udp endpoint.

bear454: UDP endpoint creation is clearly _intended_ to be supported by the SDK, and is supported by the ASM REST API (so many acronyms!) ; I don't see any reason to drop it; if anything we should weigh in on getting the SDK fixed up.

smarlowucf: Yes, I did not add the full explanation.

As far as azurectl endpoint creation UDP is unavailable. An attempt to create a UDP port returns error due to the IdleTimeoutInMinutes being [defaulted to 4](https://github.com/SUSE/azurectl/blob/master/azurectl/commands/compute_endpoint.py#L143).

This however matches the [ConfigurationSetInputEndpoint](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-servicemanagement-legacy/azure/servicemanagement/models.py#L1189) constructor in the SDK which also defaults to 4.

The [serializer](https://github.com/Azure/azure-sdk-for-python/blob/master/azure-servicemanagement-legacy/azure/servicemanagement/_serialization.py#L851) for SDK also always includes the value independent of protocol.

So this is something that could be fixed in SDK and azurectl but I don't know what the timeline is for Microsoft supporting these legacy endpoints in the SDK.

rjschwei: The way things look based on the latest information available today we are probably going to need the ASM implementation for at least another year. ARM still has no publishing support. Given that there is a large long tail of ASM instances and easy migration from ASM to ARM has not been made public I'd say it will be more than 1 year. Thus I think we should consider helping on the SDK side to get this to work.

rjschwei: You can file a new issue for azurectl indicating that the udp option is not working, but we removing it is IMHO not the best approach to address the issue.

smarlowucf: Okay, so as far as this update method it will remain as is and I will create an issue for udp and look into that on both ends (azurectl/SDK).

azurectl/commands/compute_endpoint.py
```diff
@@ -23,6 +23,13 @@
            [--idle-timeout=<minutes>]
            [--udp]
            [--wait]
+       azurectl compute endpoint update --cloud-service-name=<name> --name=<name>
```
bear454: This command should be after list & show, before delete, according to https://github.com/suse/azurectl#constructing-the-docstring

azurectl/commands/compute_endpoint.py
```diff
@@ -35,6 +42,8 @@
 commands:
     create
         add a new endpoint
+    update
```
bear454: The commands section should be sorted alphabetically, according to https://github.com/suse/azurectl#constructing-the-docstring

azurectl/instance/endpoint.py
```diff
@@ -115,6 +116,64 @@ def create(
...
+            config = self.__get_network_config_for_role(role)
+
+            if not config.input_endpoints:
+                raise AzureEndpointUpdateError(
+                    "No endpoints found."
+                )
+
+            for index, endpoint in \
+                    enumerate(config.input_endpoints.input_endpoints):
+                if endpoint.name == name:
```
bear454: How do we want to handle case sensitivity? 

The API is not case sensitive, e.g. you cannot create an endpoint named 'ssh' if you have one named 'SSH'...

doc/man/azurectl::compute::endpoint.md
```diff
@@ -17,6 +17,15 @@ __azurectl__ compute endpoint create --cloud-service-name=*name* --name=*name*
     [--udp]
     [--wait]
 
+__azurectl__ compute endpoint update --cloud-service-name=*name* --name=*name*
```
bear454: Again, need to move this into CRUD order.

doc/man/azurectl::compute::endpoint.md
```diff
@@ -39,6 +48,10 @@ a port on the specified virtual machine instance. If the virtual machine's
 __instance-name__ is the same as the __cloud-service-name__, the
 __--instance-name__ argument may be omitted.
 
+## __update__
```
bear454: ... and this in alphabetical order ;-)

azurectl/instance/endpoint.py
smarlowucf: Good catch. This will be the same as the delete and show methods. We can compare the uppercase for each.

PR #213, opened 2017-03-15
Added azurectl compute vm status

schaefi:
Allows to retrieve the current status of the virtual machine.
In addition it is possible to wait until the machine is in
Running state

### DIFF
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -39,6 +39,8 @@ usage: azurectl compute vm -h | --help
        azurectl compute vm start --cloud-service-name=<name>
            [--instance-name=<name>]
            [--wait]
+       azurectl compute vm status --cloud-service-name=<name>
+           [--instance-name=<name>]
        azurectl compute vm types
        azurectl compute vm delete --cloud-service-name=<name>
            [--instance-name=<name>]
@@ -66,6 +68,9 @@ commands:
         shuts down virtual machine instance
     start
         starts the virtual machine instance
+    status
+        Retrieves status information about the current state of the
+        virtual machine
     types
         list available virtual machine types
 
@@ -111,9 +116,10 @@ options:
     --user=<user>
         user name for login, by default set to: azureuser
     --wait
-        wait for the request to succeed
+        wait for the command to enter the requested state
 """
 import os
+import time
 # project
 from base import CliTask
 from azurectl.account.service import AzureAccount
@@ -170,6 +176,8 @@ class ComputeVmTask(CliTask):
                 self.__shutdown_instance()
             elif self.command_args['start']:
                 self.__start_instance()
+            elif self.command_args['status']:
+                self.__operate_on_instance_state()
 
     def __help(self):
         if self.command_args['help']:
@@ -220,9 +228,9 @@ class ComputeVmTask(CliTask):
             reserved_ip_name=self.command_args['--reserved-ip-name']
         )
         if self.command_args['--wait']:
-            # we might want to change this to wait until the VM becomes
-            # accessible to the user
-            self.request_wait(request_id)
+            self.__get_instance_state(
+                requested_state='ReadyRole', wait=True
+            )
         self.result.add(
             'instance',
             request_id
@@ -248,7 +256,10 @@ class ComputeVmTask(CliTask):
             cloud_service, complete_deletion
         )
         if self.command_args['--wait']:
-            self.request_wait(request_id)
+            # deletion is done when no instance state exists anymore
+            self.__get_instance_state(
+                requested_state='Undefined', wait=True
+            )
         self.result.add(
             'cloud_service:' + cloud_service,
             request_id
@@ -262,7 +273,10 @@ class ComputeVmTask(CliTask):
             instance_name
         )
         if self.command_args['--wait']:
-            self.request_wait(request_id)
+            # deletion is done when no instance state exists anymore
+            self.__get_instance_state(
+                requested_state='Undefined', wait=True
+            )
         self.result.add(
             'instance:' + instance_name,
             request_id
@@ -278,7 +292,8 @@ class ComputeVmTask(CliTask):
             instance_name
         )
         if self.command_args['--wait']:
-            self.request_wait(request_id)
+            # FIXME: This is a problem, the state did not change during reboot
+            pass
         self.result.add(
             'reboot:' + instance_name,
             request_id
@@ -295,7 +310,13 @@ class ComputeVmTask(CliTask):
             self.command_args['--deallocate-resources']
         )
         if self.command_args['--wait']:
-            self.request_wait(request_id)
+            if self.command_args['--deallocate-resources']:
+                wait_state = 'StoppedDeallocated'
+            else:
+                wait_state = 'Stopped'
+            self.__get_instance_state(
+                requested_state=wait_state, wait=True
+            )
         self.result.add(
             'shutdown:' + instance_name,
             request_id
@@ -311,10 +332,38 @@ class ComputeVmTask(CliTask):
             instance_name
         )
         if self.command_args['--wait']:
-            self.request_wait(request_id)
+            self.__get_instance_state(
+                requested_state='ReadyRole', wait=True
+            )
         self.result.add(
             'start:' + instance_name,
             request_id
+        )
+
+    def __get_instance_state(self, requested_state=None, wait=False):
+        instance_name = self.command_args['--instance-name']
+        if not instance_name:
+            instance_name = self.command_args['--cloud-service-name']
+        status = self.vm.instance_status(
+            self.command_args['--cloud-service-name'],
+            instance_name
+        )
+        if requested_state and wait:
+            request_timeout = 5
+            while status != requested_state:
+                time.sleep(request_timeout)
+                status = self.vm.instance_status(
+                    self.command_args['--cloud-service-name'],
+                    instance_name
+                )
+        return status
+
+    def __operate_on_instance_state(self):
+        instance_name = self.command_args['--instance-name']
+        if not instance_name:
+            instance_name = self.command_args['--cloud-service-name']
+        self.result.add(
+            'status:' + instance_name, self.__get_instance_state()
         )
         self.out.display()
 
--- a/doc/man/azurectl::compute::vm.md
+++ b/doc/man/azurectl::compute::vm.md
@@ -36,6 +36,9 @@ __azurectl__ compute vm start --cloud-service-name=*name*
     [--instance-name=name]
     [--wait]
 
+__azurectl__ compute vm status --cloud-service-name=*name*
+    [--instance-name=<name>]
+
 __azurectl__ compute vm types
 
 __azurectl__ compute vm delete --cloud-service-name=*name*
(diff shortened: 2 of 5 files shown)


schaefi:
[azurectl - Next+1](https://trello.com/c/TDNWTWZW/167-azurectl-next-1)

azurectl/commands/compute_vm.py
```diff
@@ -107,8 +113,11 @@
         user name for login, by default set to: azureuser
     --wait
         wait for the request to succeed
+    --wait-until-running
+        waits until the virtual machine state has changed to: Running
```
rjschwei: I think it would be nice to just make "wait" take an argument instead of adding another options, thus "--wait status" or "--wait running"

doc/man/azurectl::compute::vm.md
```diff
@@ -136,3 +144,7 @@ User name for login. Defaults to: __azureuser__
 ## __--wait__
 
 wait for the request to change its status to succeeded
+
+## __--wait-until-running__
+
+waits until the virtual machine state has changed to: __Running__
```
rjschwei: Dito, I think combining --wait and --wait-util-running is the way to go.

schaefi: ok that can be done, stay tuned

schaefi:
While implementing the change in --wait I went concerned if it was a good idea for the following reasons:

1. an option value is now required, change in behavior
2. all existing wait options except for status can only wait for the request to complete, there is no other choice. Thus an option with a required value when only one is possible feels odd

rjschwei:
On 03/15/2017 11:50 AM, Marcus Schäfer wrote:
> While implementing the change in --wait I went concerned if it was a good idea for the following reasons:
> 
> 1. an option value is now required, change in behavior

Can be addressed by making "status" the default thus "--wait" == "--wait
status"

> 2. all existing wait options except for status can only wait for the request to complete, there is no other choice. Thus an option with a required value when only one is possible feels odd
> 
> 

would also be addressed by making the "status" value the default

-- 
Distinguished Architect                       LINUX
Team Lead Public Cloud
rjschwei@suse.com
IRC: robjo

schaefi:
That's by intention not possible with docopt. You can't have a --wait and a --wait=<value>. Either one or the other variant is possible. You can have default values for an option but those apply only if that option is not supplied.

smarlowucf:
@schaefi Can they be mutually exclusive?

```[--wait | --wait=<qualifier>]```

**Comment by schaefi** [comment]
nope I have tried all that, the docopt people intentionally implemented it that way, an option must be unique

**Comment by rjschwei** [comment]
OK, that sucks, so we are stuck with the awkward "--wait status" for commands that have no second value or with two command line args :( I don't like that behavior. I tend to lean toward leaving "--wait status", i.e. the current implementation, but if you feel multiple arguments are preferable I am OK with it.

**Comment by schaefi** [comment]
Ok, then let's go with the current implementation

**Comment by schaefi** [comment]
Before I merge this I have a few other questions:

1. @bear454 if we go with the --wait option change, thus this affect other code parts or documentation where we use azurectl

2. @bear454  I saw in the smt script the retrieval of the VM state is done with a different api request (role request) The way I solved it also worked for me, however are there any objections from your side ?

3. Does a change in an option behavior cause a major or minor version update regarding our policy ? I think yes as we would provide an incompatibility to the former version

Thanks

**Comment by rjschwei** [comment]
Yes, major version bump, command line is incompatible with previous version

**Comment by bear454** [comment]
@schaefi :

1. Why change `--wait` argument... IMO this is just 'fixing' it - I don't see any value in waiting for the request_id when the _intent_ is to start a VM.

2. That's insufficient - having the deployment 'Running' means the cloud service is available; we need `instance_status == 'ReadyRole'` on the applicable role; that's when the VM is actually finished provisioning.

**Comment by schaefi** [comment]
>     1. Why change --wait argument... IMO this is just 'fixing' it - I
>        don't see any value in waiting for the request_id when the intent
>        is to start a VM.

Don't understand what you mean, the change would change --wait from a bool
option into one with a value. As this can't be done in docopt without
introducing an incompatible change... I asked

>     2. That's insufficient - having the deployment 'Running' means the
>        cloud service is available; we need instance_status == 'ReadyRole'
>        on the applicable role; that's when the VM is actually finished
>        provisioning.

The code gets the deployment with instance information. From there it
checks the instance in the deployment and its status. The result looks
the same to me as the role check. During provisioning my code returns
'Initializing', after that 'Running', when stopped I got 'Stopped'.
I see the same satus message as I see from the web api

**Comment by schaefi** [comment]
Ok, I'd like to move forward on this also with regards to the other open PR. I reverted the incompatible wait option change and will go with an extra option as it was before

**Comment by schaefi** [comment]
There is then only one open question which of the following code snippets needs to be preferred or if the way I get the status is insufficient:

```python
properties = self.service.get_hosted_service_properties(instance_name, embed_detail=True)
for deployment in properties.deployments:
    if deployment.name == instance_name:
        instance_state = deployment.status
```

vs.

```python
properties = self.service.get_properties(instance_name)
for role in cloud_service_properties['roles']:
    if (role['name'] == new_instance_name):
        instance_state = role['instance_status']
```

Thanks

bear454:
I'm sorry if my take on the wait options wasn't clear... I'll try again:

For me, this is fundamentally a question of what the `--wait` argument means. I don't agree that `--wait` means "return, when the asynchronous request, identified by an ID returned by an API call, has a completed status, of either success or failure". That's too technical a decision, and too dependent on the implementation details. (It also assumes a lot of domain knowledge of the Azure REST API).

For me, the `--wait` argument means "return, when the action I intended performed is complete." Wait until the disk is attached. Wait until the endpoint removed. Wait until the VM is created.

That being said, the _current_ implementation of `--wait` for `compute vm create` is flawed, in that it does not wait until the VM is created. I appreciate you extending the wait operation to better meet the simple expection of "wait until the VM is created", but I don't see any value in identifying, or sorting out the implementation details behind it, by saying "wait until the asynchronous request is dequeued". 

There should only be the one behavior of `--wait`. We had a _bug_ in `--wait` on `compute vm create`. You've (potentially) fixed that bug. Don't introduce a new one by complicating the UI.

schaefi:
ok got it, actually I had just planned a small change, anyway I changed the wait behavior to really wait for the command to complete. Complete means:

* vm create
   wait for Running state

* vm reboot
  wait for state change from Initializing back to Running (I haven't tested it, needs testing)

* vm shutdown
  wait for Stopped state (I haven't tested it, needs testing)

* vm delete
  wait for Undefined state, all gone

* vm status
  request state at call time

My tests are not finished, so please don't merge. Of course any review or clear mistake in my state change thinking is much appreciated

schaefi:
I have tested the state and state changes with a real instance. It has turned out that only the role status really changes, thus I changed my code to use the roles information like James did in the smtReplace code.

So far so good however there is still one problem. If an instance is rebooted there is no change in any of the states. Thus I don't see a way to detect or wait for a reboot. You can see this in the FIXME comment in the code

Thoughts ?

schaefi:
waiting for vm create, delete and shutdown works nicely

rjschwei:
Well, if we redefine --wait as we are doing now we still need a major version bump. I am not sure that the change from "asynchronous wait" to "serialized wait" while taking away the asynchronous wait option complete is the step we want to take. One can probably construct a reasonable case where asynchronous wait is sufficient in a given script.

I think having wait be an option rather than a switch is appropriate. We can use generic options such as 

--wait request  -> asynchronous wait, return when request is processed by Azure
--wait complete -> synchronous wait, return when the state of the operation changes to complete, VM running, disk attached or detached etc.

schaefi:
With regards to VM operations I couldn't think of a good use case when waiting for the request to be processed would make any sense. I basically agree to what James said. From a user perspective if you want to wait for the VM to be created, started, rebooted or shut down you always want to continue when this state is reached and not until Azure has handled the request.

That's why I would agree to the statement that I did wrong from the beginning on, and this code change just fixes what we should have done from the first implementation ;) I'm ok with a major version bump but I don't think we need to over complicate the wait option

We need a solution how to wait for the reboot though, if possible

rjschwei:
OK, lets focus on the vm and lets look for guidance about what others have done.

aws cli - returns when the vm is started and does not wait until one can ssh in
gcloud - waits until the vm is indicated as running, but this state does not coincide with ability to ssh in

I have not yet looked at the az tools.

What @bear454 is advocating is waiting for a state change, which is a distinctly different opperation. While I agree that this has the appeal of simplicity and presumed intent it is also, IMHO over reaching and imposing "what works for me" on the rest of the user base. There are certainly situations where I don't want to wait until the VM is up and running. What if I want to create and attach a disk in s script I I start a vm? I should not have to wait until the vm state is "Running" before my script goes on and creates a disk and then attaches it.

bear454:
@rjschwei said
>  There are certainly situations where I don't want to wait until the VM is up and running. What if I want to create and attach a disk in s script I I start a vm? I should not have to wait until the vm state is "Running" before my script goes on and creates a disk and then attaches it.

You can't add a disk to the VM until after the agent finishing provisioning; while you may execute the request, it won't be completed until the agent signals provisioning to be complete... which practically synchronous to when the vm is ready. (Keep in mind, "ready" only implies SSH access from _our_ perspective - another system could have another behavior).

Your proverbial disk attach will not be completed until provisioning completes, and the instance hits 'ReadyRole'; only then will further actions on the instance be dequeued and performed.

schaefi:
right, before the ReadyRole I would not expect any action to the VM to be successful. I also think it really does not hurt to wait until the system is responsive before performing any action. Robert if a wait guarantees me I can work with the system I would not complain if one or the other action could have happened 1sec before if we would have written the wait differently. I don't understand why you are so strict on the wait ? if we leave it as it is now, wait is useless, if we go with the current solution wait is finally useful, if we change wait from a bool option to some with a value we confuse our users and I bed the first question will be why we made wait so complex

rjschwei:
Thanks for the additional clarifications.

schaefi:
Ahm, now you merged code with a FIXME statement and a nop for wait on reboot. This is a little bit unfortunate as we were still discussing how we, if we should; wait for the reboot

...well too late

rjschwei:
On 03/24/2017 10:37 AM, Marcus Schäfer wrote:
> Ahm, now you merged code with a FIXME statement and a nop for wait on reboot. This is a little bit unfortunate as we were still discussing how we, if we should; wait for the reboot
> 
> ...well too late
> 
Sorry, I though we were done with the review, my bad I should have
double checked.

smarlowucf:
> So far so good however there is still one problem. If an instance is rebooted there is no change in any of the states. Thus I don't see a way to detect or wait for a reboot. You can see this in the FIXME comment in the code

@schaefi This is similar across the other providers (GCE, EC2). State doesn't change on a soft reboot. Don't think there's a way to detect a soft reboot has finished aside from establishing an SSH connection. Ran into this issue with the testing framework. Had to force hard reboot and stop sshd prior to shutdown to prevent SSH connection prior to reboot actually initiating.

IMHO the best option is to keep it consistent with the other waits and just wait for 'ReadyRole' similar to `__start_instance` even though it doesn't change.

PR #217, opened 2017-03-21
Add azurectl compute vm start.

smarlowucf:


schaefi:
ok, I'd like to move on and get your stuff in, I have updated the other PR in a way that we could merge without conflicts. Let's get your stuff in it looks great :)

PR #218, opened 2017-03-21
Fix mutable default values.

smarlowucf:
- Default values should not contain mutable objects such as lists.
- Default lists are created once at function definition instead of with each call.


smarlowucf:
@bear454 Would it be safe to remove the default for the validations.py method as well? The keys are required for the method to perform the given task.

schaefi:
Thanks, I wonder why the code checker did not complain earlier on this part

PR #219, opened 2017-03-31
Follow up fix for vm wait states

schaefi:
On reboot no state change can be determined from the API.
Thus we define the safe default and wait for the ReadyRole
even though the vm might already be in that state before
reboot

### DIFF
--- a/azurectl/commands/compute_vm.py
+++ b/azurectl/commands/compute_vm.py
@@ -292,8 +292,12 @@ class ComputeVmTask(CliTask):
             instance_name
         )
         if self.command_args['--wait']:
-            # FIXME: This is a problem, the state did not change during reboot
-            pass
+            # On reboot the state of the VM did not change in the API
+            # Thus there is no real chance to wait for the reboot to
+            # complete. The safe default is to wait for the ready role
+            self.__get_instance_state(
+                requested_state='ReadyRole', wait=True
+            )
         self.result.add(
             'reboot:' + instance_name,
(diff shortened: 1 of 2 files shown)


schaefi:
In reply to information from @smarlowucf 

> This is similar across the other providers (GCE, EC2). State doesn't change on a soft reboot. Don't  
> think there's a way to detect a soft reboot has finished aside from establishing an SSH connection.
> Ran into this issue with the testing framework. Had to force hard reboot and stop sshd prior to
> shutdown to prevent SSH  connection prior to reboot actually initiating.

> IMHO the best option is to keep it consistent with the other waits and  just wait for 'ReadyRole' similar 
> to __start_instance even though it doesn't change.

I agree to let's wait for the ready role on reboot

azurectl/commands/compute_vm.py
```diff
@@ -292,8 +292,12 @@ def __reboot_instance(self):
             instance_name
         )
         if self.command_args['--wait']:
-            # FIXME: This is a problem, the state did not change during reboot
-            pass
+            # On reboot the state of the VM did not change in the API
```
rjschwei: Please modify the comment as "soft reboot" to be more specific and differentiat a reboot initiated by the framework (stop-start) vs a reboot initiated from within the vm, i.e. shutdown -r

schaefi:
@rjschwei could you check again if the changes meets your request ? Thanks

PR #220, opened 2017-06-30
Update requirements

smarlowucf:
Had an issue building due to missing pyOpenSSL requirement in setup. In reviewing all the requirements I came across some other things as well:

  - dnspython doesn't appear to be a requirement (build and tests pass without it, did not find any imports in searching the codebase and nothing in dependency tree)
  - Don't think setuptools should be required aside from rpm build
  - Requests doesn't appear to be a direct requirement (required by azure storage and servicemanagement legacy)
    - This is okay as requirements files can contain the entire dependency tree however this is the only indirect dependency thus it's a bit confusing

Thoughts on these and also a version to pin pyOpenSSL to in setup?

### DIFF
--- a/.travis.requirements.txt
+++ b/.travis.requirements.txt
@@ -15,8 +15,6 @@ APScheduler
 pyliblzma
 coverage
 pyOpenSSL
-requests
-dnspython
 pytest
 pytest-cov
 azure-storage
--- a/.virtualenv.dev-requirements.txt
+++ b/.virtualenv.dev-requirements.txt
@@ -11,7 +11,6 @@
 
 -r .virtualenv.requirements.txt
 
-setuptools
 pytest
 pytest-cov
 mock
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -16,9 +16,7 @@ pyliblzma
 azure-storage
 azure-servicemanagement-legacy
 python-dateutil
-dnspython
 
 pyOpenSSL
-requests
-
+setuptools
 future
--- a/package/spec-template
+++ b/package/spec-template
@@ -18,7 +18,7 @@ Requires:       python-pyliblzma
 Requires:       python-azure-sdk >= 1.0.3
 Requires:       python-azure-sdk-storage >= 0.30.0
 Requires:       python-dateutil
-Requires:       python-dnspython
+Requires:       python-pyOpenSSL
 Requires:       python-setuptools
 Requires:       python-future
 Requires:       man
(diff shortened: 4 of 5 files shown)


smarlowucf:
Also, I just noticed the travis requirements file has azure-storage but is missing azure-servicemanagement-legacy.

schaefi:
right the sdk is taken from the git because of the strange pip setup of the azure sdk. However the data on pip has changed and we are also currently changing the packaging. Thus it might be that we need to adapt the source we need for azurectl in the near future. Hopefully to just depend on the azure-servicemanagement-legacy module. But let us handle that once we finished the packaging and don't confuse ourselves at several ends :)

schaefi:
> dnspython doesn't appear to be a requirement 

agreed

schaefi:
> Requests doesn't appear to be a direct requirement

agreed

schaefi:
> Thoughts on these and also a version to pin pyOpenSSL to in setup

I haven't seen a minimum required version. On the distros I have tested this the version was always
sufficient. Thus it's imho ok to leave it as you've added it. I wonder why the missing requirement was
not observed earlier, thanks

schaefi:
> Don't think setuptools should be required aside from rpm build

but setuptools is required also at runtime because of the entry point code it generated and thus depends on it. I think a more clean solution would be to:

* add setuptools to .virtualenv.requirements.txt
* delete setuptools from .virtualenv.dev-requirements.txt because it inherits from .virtualenv.requirements.txt
* change the README.md lines

    ```
    $ pip2.7 install -U pip setuptools
    $ pip2.7 install -r .virtualenv.dev-requirements.txt
    ```

    to

    ```
    $ pip install -U pip
    $ pip install -r .virtualenv.dev-requirements.txt
    ```

smarlowucf:
> but setuptools is required also at runtime because of the entry point code it generated and thus depends on it. I think a more clean solution would be to:
> 
> add setuptools to .virtualenv.requirements.txt
> 
> delete setuptools from .virtualenv.dev-requirements.txt because it inherits from .virtualenv.requirements.txt [alreay done by you]
> 
> change the README.md lines

I mistakenly thought setuptools was part of standard library but it is rather included with certain distributions of Python along with pip.

Added the requirement back and moved from dev to general requirements file. I think it's okay to leave in readme as this will ensure it's up-to-date.

smarlowucf:
> I haven't seen a minimum required version. On the distros I have tested this the version was always
> sufficient. Thus it's imho ok to leave it as you've added it. I wonder why the missing requirement was
> not observed earlier, thanks

I noticed the missing requirement trying to configure a tox testing environment. (This uses pip/venv/setup.py etc) The spec does not run tests so I don't think missing requirement would block build.

schaefi:
:+1:

PR #221, opened 2017-07-10
2to3

schaefi:
Move azurectl code base to python3


schaefi:
I would say it's impossible to review that, thus please give it a real test

the most intrusive change was the move from pyliblzma which was not ported to py3 because of lzma in py3 core. This impacts the upload code of xz compressed data to Azure

test/unit/account_setup_test.py
```diff
@@ -257,7 +257,7 @@ def test_add_first_region_as_default(self):
         )
         assert setup.list()['account_region_map']['region'] == 'region:earth'
 
-    @patch('__builtin__.open')
+    @patch_open
```
smarlowucf: If this is Python3 only can this be simplified to just patch builtins?

test/unit/app_test.py
```diff
@@ -1,7 +1,7 @@
 import sys
 import pytest
 import mock
-from test_helper import *
+from .test_helper import raises
```
smarlowucf: Unrelated to the Python3 move so I can open a separate enhancement request but if pytest is used for testing then can the [raises helper](https://docs.pytest.org/en/latest/builtin.html#helpers-for-assertions-about-exceptions-warnings) be used in all these tests?

Using context manager around the very specific statement that throws the error is a stronger test imho. It also allows for comparing the exception message wording.

azurectl/utils/xz.py
```diff
@@ -34,40 +34,30 @@ def __init__(self, lzma_stream, buffer_size=LZMA_STREAM_BUFFER_SIZE):
         self.lzma_stream = lzma_stream
         self.buffer_size = int(buffer_size)
         self.lzma = lzma.LZMADecompressor()
-        self.finished = False
 
     def read(self, size):
```
smarlowucf: Can this module be dropped and switched to use Python3 lzma library directly which [supports xz compression](https://docs.python.org/3/library/lzma.html#examples)?

smarlowucf: Ah, getting the uncompressed size would be an issue. But maybe this can be simplified to just a wrapper of lzma? And simplify the size method to something like...

def uncompressed_size(f):
    f.seek(0, os.SEEK_END)
    size = f.tell()
    return size

test/unit/account_setup_test.py
schaefi: I prefer to keep the amount of changes to make the code eventually py2 and py3 compatible as small as possible. The open patching goes into that direction

azurectl/utils/xz.py
schaefi: I am using lzma from py3 core, as I wrote in my initial comment pyliblzma does not exist for py3
You have a point here with the uncompressed_size, maybe we can get rid of calling xz. I'll check that. thanks

test/unit/app_test.py
schaefi: Well yes that's true, I just thought moving 144 raise tests to context manager might be a waste of time for a project which is not expected to grow much anymore ;) From that perspective there would be even more which could and should change, e.g the use of __methods vs. _methods and more. Let me check how much effort it would be to make things nicer.

However first and foremost I think we should test if the code actually really works as expected which was the reason why I asked for some real life tests

azurectl/utils/xz.py
schaefi: ok done

test/unit/account_setup_test.py
schaefi: nevertheless I'll remove it with a builtins patch

package/spec-template
```diff
@@ -1,26 +1,26 @@
 #
 # spec file for package python-azurectl
 #
-Name:           python-azurectl
+Name:           python3-azurectl
```
rjschwei: I know technically we should call this "python3-azurectl". However, if we rename the package we are going to run into a  lot of procedural problems in SLE 12 and SLE 15.

One is that packages cannot be removed, which means we cannot get rid of python-azurectl in SLE 12 and then we'd have python-azurectl and python3-azurectl and people will get the impression that they are equal and one can use either the Py2 or Py3 version. However we will only continue to develop in Py3. Further I suspect that in SLE 15 py3 will be "python" thus we'd have to switch back to "python-azurectl".

IMHO we should break the naming convention now in order to avoid flip flopping the name as we look ahead to SLE 15. Also, given that azurectl is primarily a command line tool it is really immaterial to the user if the py3 or py2 interpreter runs.

package/spec-template
```diff
@@ -1,26 +1,26 @@
...
-Requires:       python-pyOpenSSL
-Requires:       python-setuptools
-Requires:       python-future
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+Requires:       python3-docopt
+Requires:       python3-APScheduler
+Requires:       python3-azure-sdk >= 1.0.3
+Requires:       python3-azure-sdk-storage >= 0.30.0
+Requires:       python3-dateutil
```
rjschwei: python3-python-dateutil (although the python-python-datutil rename may not have made it to SLE yet)

package/spec-template
```diff
@@ -1,26 +1,26 @@
...
-Requires:       python-azure-sdk-storage >= 0.30.0
-Requires:       python-dateutil
-Requires:       python-pyOpenSSL
-Requires:       python-setuptools
-Requires:       python-future
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+Requires:       python3-docopt
+Requires:       python3-APScheduler
+Requires:       python3-azure-sdk >= 1.0.3
```
rjschwei: I think you can break this into components already with this commit, rather than doing a 2to3 commit followed by a commit that changes the dependencies.

package/spec-template
```diff
@@ -1,26 +1,26 @@
...
+Requires:       python3-azure-sdk-storage >= 0.30.0
+Requires:       python3-dateutil
+Requires:       python3-pyOpenSSL
+Requires:       python3-setuptools
+Requires:       python3-future
+Requires:       python3-six
```
rjschwei: We should not need -future and -six since we are switching to py3 only

package/spec-template
schaefi: I had that before and thinking the same way, problem is that the buildservice auto checks complains about it, not sure if people will accept it when we submit

package/spec-template
schaefi: yes that was just a left over... will fix

package/spec-template
schaefi: I thought the python3-azure-sdk will stay as a meta package with a version bump and all we need to do here is to change the min required version for the azure-sdk package ?

schaefi: I think we had the same discussion during your stay in Nue but I can't remember the result of the discussion. So let's write the preferred solution down here :) ETOOMANYBUILDINGSITES :))

rjschwei: Let me check with dimstar. Since the devel project for python-azurectl is Cloud:Tools and not d:l:p we do not quite have the same scruteny :smile:

schaefi:
Please find package builds of this branch here:

* https://build.opensuse.org/project/show/home:sax2:branches:Cloud:Tools

package/spec-template
schaefi: yep, let me know what dimstar sayed, thanks

rjschwei: Yes python-azure-sdk will become a meta package in SLE 12, but the intention is to drop the meta package for SLE 15. Because we want to drop python-azure-sdk in SLE 15 and we don't want to maintain forked versions of azurectl it is, IMHO easier to break the dependencies of azurectl down to the SDK component level now.

schaefi: ah yes sle15 was the reason, I remember, ok thanks

smarlowucf:
In testing found this issue:

```
>       fingerprint = fingerprint.split('=')[1]
E       TypeError: a bytes-like object is required, not 'str'

.tox/py36/lib/python3.6/site-packages/azurectl/instance/cloud_service.py:82: TypeError
```

Pushed an update to this branch.

schaefi:
@smarlowucf Thanks for helping, but it's easier for me to handle additional contributions to this branch if you create your own branch based on this one and send a pull request which I can simply merge to this one when completed. Hope you don't mind but I will force push the history now

schaefi:
Within the scope of my testing the port to py3 works now. Please give me your approval if you think we are ready to merge. Thanks

rjschwei:
+1 for the code, we still have to sort out the package naming, thus please do not merge yet.

schaefi:
ok

schaefi:
blocked on spec file decisions

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.3.2
+current_version = 2.4.0
```
bear454: I'd argue that we need to go to 3.0.0; this is more accurately described as an "incompatible API change" than "add[ing] functionality in a backwards-compatible manner", as we're clearly not going to be backwards-compatible.

README.md
```diff
@@ -247,33 +247,26 @@ for the desired Python version.
```
bear454: Need to change L243 as well "azurectl is compatible with Python 2.7.x and greater"

azurectl/azurectl.py
```diff
@@ -15,10 +15,10 @@
 import docopt
 
 # project
-import logger
-from app import App
-from defaults import Defaults
-from azurectl_exceptions import AzureError
+from . import logger
```
bear454: Using local imports has caused us trouble in the past, when using azurectl as a library.

azurectl/commands/compute_data-disk.py
```diff
@@ -99,7 +99,7 @@
 # project
 from azurectl.logger import log
 from azurectl.account.service import AzureAccount
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_endpoint.py
```diff
@@ -79,7 +79,7 @@
         wait for the request to succeed
 """
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_image.py
```diff
@@ -127,7 +127,7 @@
 
 # project
 from azurectl.logger import log
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_request.py
```diff
@@ -32,7 +32,7 @@
         request id number returned from azurectl for asynchronous request
 """
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_reserved-ip.py
```diff
@@ -53,7 +53,7 @@
         wait for the request to succeed
 """
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_shell.py
```diff
@@ -24,7 +24,7 @@
 from textwrap import dedent
 
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/compute_vm.py
```diff
@@ -121,7 +121,7 @@
 import os
 import time
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/setup_account.py
```diff
@@ -80,7 +80,7 @@
 import os
 
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/storage_account.py
```diff
@@ -96,7 +96,7 @@
 import string
 
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/storage_container.py
```diff
@@ -65,7 +65,7 @@
 import datetime
 
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/storage_disk.py
```diff
@@ -70,7 +70,7 @@
 from apscheduler.schedulers.background import BackgroundScheduler
 
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/commands/storage_share.py
```diff
@@ -36,7 +36,7 @@
         name of the files share
 """
 # project
-from base import CliTask
+from .base import CliTask
```
bear454: `from azurectl.commands.base import CliTask` ?

azurectl/help.py
```diff
@@ -13,7 +13,7 @@
 #
 import subprocess
 # project
-from azurectl_exceptions import AzureHelpNoCommandGiven
+from .azurectl_exceptions import AzureHelpNoCommandGiven
```
bear454: `from azurectl.azurectl_exceptions import AzureHelpNoCommandGiven` ?

azurectl/instance/reserved_ip.py
```diff
@@ -12,6 +12,7 @@
 # limitations under the License.
 #
 # project
+from azurectl.defaults import Defaults
```
bear454: :+1:

azurectl/version.py
```diff
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '2.3.2'
+__VERSION__ = '2.4.0'
```
bear454: Again, not sold on the version number, re semantic versioning.

rjschwei: Agreed, it should be 3.0

bear454:
@schaefi do you want me to submit a PR with the relative imports fixed?

azurectl/azurectl.py
smarlowucf: These should be okay in Python3. The period is relative import rather than local. Although I agree absolute imports are more explicit.

schaefi:
@bear454 sorry for the delay, now that the packaging questions seems to have answered we should also go through the code and fix (prevent) the relative imports yes 👍 This is currently a mix and I'm fine if you have a free slot to send a pr on that one. If not just let me know then I'll do it next. Thanks much

smarlowucf:
@schaefi 

> we should also go through the code and fix (prevent) the relative imports yes

I tested azurectl from code and the imports were okay since they were updated to explicit relative with the periods instead of implicit. However, it is recommended to use absolute imports so I went ahead and made the pr.

schaefi:
yay ready to merge :)

PR #222, opened 2017-07-11
- Python 2 to Python 3 conversion

rjschwei:
+ We are switching to Python 3, thus we do not need to depend on
    the future module


schaefi:
Thanks 👍

PR #223, opened 2017-07-12
- Explicitly convert string to int for comparison operation

rjschwei:
+ In Python 3 the string needs to be explicitly converted to
    an int to be used in the comparison operation


smarlowucf:
Hit the same problem, the request id's need conversion from hex or just a string if check...

```if cloud_service_request_id: # assert not an empty string or None```

or

```if int(cloud_service_request_id, 16) > 0:```

**Comment by rjschwei** [comment]
Truth testing has not changed AFAIK

Python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = 'a'
>>> if a:
...   print('Worked')
... 
Worked

**Comment by smarlowucf** [comment]
Sorry didn't mention the error but only possible solutions...

With this PR I get the following error because the request ID is really a string not an int.

```
ERROR: Unexpected error:
  File "/home/smarlow/projects/venvs/azurectl3/bin/azurectl", line 11, in <module>
    load_entry_point('azurectl', 'console_scripts', 'azurectl')()
  File "/home/smarlow/projects/azurectl/azurectl/azurectl.py", line 45, in main
    App()
  File "/home/smarlow/projects/azurectl/azurectl/app.py", line 31, in __init__
    task_class().process()
  File "/home/smarlow/projects/azurectl/azurectl/commands/compute_vm.py", line 166, in process
    self.__create_cloud_service()
  File "/home/smarlow/projects/azurectl/azurectl/commands/compute_vm.py", line 245, in __create_cloud_service
    if int(cloud_service_request_id) > 0:
ValueError: invalid literal for int() with base 10: 'e208fe07ca6006e7a3e64f6309a928b8'
```

**Comment by rjschwei** [comment]
OK, that would imply that the request_id we are after was never meant to be compared to 0 in the first place. I'll go look at what the SDK has to say about he request id

**Comment by rjschwei** [comment]
I guess I got lucky when I tested this, oh well sometimes even a blind ....

Latest push should fix the issue, this was a latent bug, should have never worked, but for the "favors" python2 played.

**Comment by rjschwei** [comment]
@schaefi the only documentation I could find about what request_id should be was here: https://docs.microsoft.com/en-us/rest/api/storageservices/troubleshooting-api-operations

This of course does not specify what the type is and in the code all I found was

def parse_response_for_async_op(response):
    ''' Extracts request id from response header. '''

    if response is None:
        return None

    result = AsynchronousOperationResult()
    if response.headers:
        for name, value in response.headers:
            if name.lower() == 'x-ms-request-id':
                result.request_id = value

    return result

Where value could of course be of any type, it appears that based on https://msdn.microsoft.com/en-us/library/azure/ee460783.aspx the request-id should be treated as a string, in which case the current implementation would probably be correct. However in the tests the request-id is treated as an int, thus the tests fail. The only thing we do know for sure is that the API does not return an int. Thus the tests do not replicate the expected value from the API and thus the SDK.

**Comment by rjschwei** [comment]
More empirical information, for "storage disk upload" and "compute image create" the "len(request_id" implementation works, but for "compute vm create" the request_id is an integer and the len() fails.

Great that the REST API is inconsistent with returned values. Then again what would
 one expect from Msft.

**Comment by schaefi** [comment]
Thanks for the investigations. With all your information I think it would be a good idea to pass along any request id to a method to allow a common validation as we need it. Thus I opened #225 and like to ask for a review. So far the method just calls format() on the value which might be not complete. But with the one place to deal with the value we can enhance the method as needed

Thoughts ?

**Comment by schaefi** [comment]
ok so this one needs a fix for the unit test to be complete

**Comment by rjschwei** [comment]
OK, finally, with this we make it through a complete cycle:

storage disk upload
compute image create
compute vm create

and the tests pass :smile:

File: azurectl/commands/compute_vm.py [hunk 1]
```diff
@@ -238,11 +239,13 @@ def __create_instance(self):
...
-            self.command_args['--cloud-service-name'],
-            self.config.get_region_name()
+        cloud_service_request_id = Defaults.unify_id(
+            self.cloud_service.create(
+                self.command_args['--cloud-service-name'],
+                self.config.get_region_name()
+            )
         )
-        if cloud_service_request_id > 0:
+        if int(cloud_service_request_id, 16) > 0:
```
schaefi: Once the superfluous unify_id call is deleted the test in test/unit/commands_compute_vm_test.py will fail again which is correct because we expect cloud_service.create() to return a unified id as a string which in this case needs a conversion to a number. Thus the test has to be adapted:

```
--- a/test/unit/commands_compute_vm_test.py
+++ b/test/unit/commands_compute_vm_test.py
@@ -29,7 +29,7 @@ class TestComputeVmTask:
         )
         cloud_service = mock.Mock()
         cloud_service.create = mock.Mock(
-            return_value=42
+            return_value='42'
```

schaefi: With our validation to the request id we handle it as a string in any case. That has an impact on the line

```
self.request_wait(cloud_service_request_id)
```

self.request_wait is causing a service.get_operation_status(self.request_id) call in the SDK code. I have checked the code and found that they always call:

```
return self._perform_get(
            '/' + self.subscription_id + '/operations/' + _str(request_id),
            Operation)
```

So the id is translated into a string which means our Defaults.unify_id() method should be safe. At least I could not find a problem in azure/servicemanagement/_common_conversion.py if the request id is already passed as a string

schaefi: Summarizing this I think we should be fine if the superfluous unify_id() call gets deleted and the test gets fixed as the pasted diff shows

azurectl/commands/compute_vm.py
```diff
@@ -238,11 +239,13 @@ def __create_instance(self):
         self.out.display()
 
     def __create_cloud_service(self):
-        cloud_service_request_id = self.cloud_service.create(
-            self.command_args['--cloud-service-name'],
-            self.config.get_region_name()
+        cloud_service_request_id = Defaults.unify_id(
```
schaefi: The return value from self.cloud_service.create() comes from the instance/cloud_service.py::create() method which already returns the id via Defaults.unify_id(). Thus the above call is superfluous and should not be needed

azurectl/commands/compute_vm.py
schaefi: Once the superfluous unify_id call is deleted the test in test/unit/commands_compute_vm_test.py will fail again which is correct because we expect cloud_service.create() to return a unified id as a string which in this case needs a conversion to a number. Thus the test has to be adapted:

```
--- a/test/unit/commands_compute_vm_test.py
+++ b/test/unit/commands_compute_vm_test.py
@@ -29,7 +29,7 @@ class TestComputeVmTask:
         )
         cloud_service = mock.Mock()
         cloud_service.create = mock.Mock(
-            return_value=42
+            return_value='42'
```

schaefi: With our validation to the request id we handle it as a string in any case. That has an impact on the line

```
self.request_wait(cloud_service_request_id)
```

self.request_wait is causing a service.get_operation_status(self.request_id) call in the SDK code. I have checked the code and found that they always call:

```
return self._perform_get(
            '/' + self.subscription_id + '/operations/' + _str(request_id),
            Operation)
```

So the id is translated into a string which means our Defaults.unify_id() method should be safe. At least I could not find a problem in azure/servicemanagement/_common_conversion.py if the request id is already passed as a string

schaefi: Summarizing this I think we should be fine if the superfluous unify_id() call gets deleted and the test gets fixed as the pasted diff shows

rjschwei:
OK, cleaned up the history and fixed the test. This should work now without superfluous calls to unify_id

schaefi:
Thanks much :+1:

PR #225, opened 2017-07-13
Added unify_id method for request id validation

schaefi:
The request id value from the Azure rest api is not consistently
of the same type which makes it hard to handle the id if it could
be a string or a number. This commit introduce a static method
called unify_id which allows us to properly format the id into
a common value so that azurectl can consume the information in
the same way all over the place. So far unify_id just formats
the information, depending on the value more python logic might
be required


rjschwei:
I think this is a good approach to deal with the problem. With this approach we can fix #223 to be int(request_id, 16) > 0 which I think was the original intend of the comparison in py2.

schaefi:
ok, so I'll merge this to the branch and we can create a follow up request for the needed changes

PR #226, opened 2017-07-25
- Switching to Python 3 makes the API "incompatible" thus we need to

rjschwei:
bump to 3.0.0
- Package is being renamed we need to use rpm accepted Provides Obsoletes
  construct


azurectl/version.py
```diff
@@ -14,4 +14,4 @@
 """
     Global version information used in azurectl and the package
 """
-__VERSION__ = '2.4.0'
+__VERSION__ = '3.0.0'
```
schaefi: why does the move to py3 make the api incompatible ? it's still the same api or am I missing the point ?
In any case please use the bumpversion tool to update the version:

```
bumpversion major
```

package/spec-template
```diff
@@ -14,7 +14,7 @@ BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
 Requires:       python3-docopt
 Requires:       python3-APScheduler
-Requires:       python3-azure-sdk-storage >= 0.30.0
+Requires:       python3-azure-storage >= 0.30.0
```
schaefi: who has done this package renaming and why ?

schaefi: to my knowledge there is no change on the azure storage side of things

rjschwei: @glaubitz submitted the package to d:l:p yesterday. The upstream name of the package on PyPi is azure-storage, thus our package should be named python3-azure-storage. I think when you originally created the python-azure-sdk-storage package it was named this way because we pulled the tarball out of the azure-sdk tarball.

Anyway, since we are trying to get all the ducks lined up and be name guideline compliant it made sense to fix this along the way.

azurectl/version.py
rjschwei: OK, used bumpversion

The API is the same per-se, agreed. However, someone running in Python 2 will notice the API disappeared, which one could argue is an incompatibility. I'd say going to 3.0.0 shows that we cooked up a major change and people shouldn't just blindly pick it up if the do not put in some effort to understand what happened.

